PR #18, opened 2017-04-16
coding style 

Ericliu001:
Just fixed a few things that Lint has picked up like redundant 'public' modifiers for interface methods and simpler generic syntax for Java7.

### DIFF
--- a/src/main/java/com/android/volley/Cache.java
+++ b/src/main/java/com/android/volley/Cache.java
@@ -28,43 +28,43 @@ public interface Cache {
      * @param key Cache key
      * @return An {@link Entry} or null in the event of a cache miss
      */
-    public Entry get(String key);
+    Entry get(String key);
 
     /**
      * Adds or replaces an entry to the cache.
      * @param key Cache key
      * @param entry Data to store and metadata for cache coherency, TTL, etc.
      */
-    public void put(String key, Entry entry);
+    void put(String key, Entry entry);
 
     /**
      * Performs any potentially long-running actions needed to initialize the cache;
      * will be called from a worker thread.
      */
-    public void initialize();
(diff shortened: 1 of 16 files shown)


src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -160,9 +160,9 @@ public void stop() {
         if (mCacheDispatcher != null) {
             mCacheDispatcher.quit();
         }
-        for (int i = 0; i < mDispatchers.length; i++) {
-            if (mDispatchers[i] != null) {
-                mDispatchers[i].quit();
+        for (final NetworkDispatcher mDispatcher : mDispatchers) {
```
jpd236: The linter is being a bit overzealous here; this is actually a violation of the "Avoid iterators" advice in the [Developing for Android](https://medium.com/google-developers/developing-for-android-ii-bb9a51f8c8b9) doc.

That being said, given that the doc acknowledges that this isn't a big deal in practice as long as it's a single allocation (that is, not being done in an inner loop), so I'm okay with it - stop() shouldn't ever be called in a loop.

jpd236:
Thanks for the cleanup!

Ericliu001:
Thank you!

src/main/java/com/android/volley/Cache.java
```diff
@@ -87,12 +87,12 @@
         public Map<String, String> responseHeaders = Collections.emptyMap();
 
         /** True if the entry is expired. */
-        public boolean isExpired() {
+        boolean isExpired() {
```
jpd236: I missed that these are methods in a class, not an interface. This was a functional change that should be reverted. (I'll take care of this)

PR #22, opened 2017-04-21
Make sure last CacheDispatcher request is GCed

### LINKED ISSUE #19: Leaking last request in CacheDispatcher
Android Studio reported a leaked `Activity` that I tracked down to `CacheDispatcher` keeping a reference to a request:
cachedispatcherqueueleak

The issue was fixed in now deprecated `mcxiaoke/android-volley` repository:
[leaking the last request object in CacheDispatcher](https://github.com/mcxiaoke/android-volley/pull/89)

And the reson for the leak is explained in:
[A small leak will sink a great ship](https://medium.com/square-corner-blog/a-small-leak-will-sink-a-great-ship-efbae00f9a0f)

tidbeck:
Fixes #19

### DIFF
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -83,11 +83,15 @@ public class CacheDispatcher extends Thread {
         // Make a blocking call to initialize the cache.
         mCache.initialize();
 
+        Request<?> request;
         while (true) {
             try {
+                // Make sure the old request will be garbage collected
+                request = null;
+
                 // Get a request from the cache triage queue, blocking until
                 // at least one is available.
-                final Request<?> request = mCacheQueue.take();
+                request = mCacheQueue.take();
                 request.addMarker("cache-queue-take");
 
                 // If the request has been canceled, don't bother dispatching it.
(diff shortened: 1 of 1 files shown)


tidbeck:
@jpd236 simplified the code and force-pushed. If you still find the code to complex, please give me some hints on how to simplify it.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -134,11 +138,12 @@ public void run() {
 
                     // Post the intermediate response back to the user and have
                     // the delivery then forward the request along to the network.
-                    mDelivery.postResponse(request, response, new Runnable() {
+                    final Request<?> finalRequest = request;
+                    mDelivery.postResponse(finalRequest, response, new Runnable() {
```
jpd236: If I understand correctly, the cause of the bug is that Dalvik won't GC a local variable of a loop even after it goes out of scope on the next iteration of that loop.

If so, by making a final variable here (that is this never nulled out), isn't this just introducing another instance of the same leak?

tidbeck:
This is not a correct fix for the issue.

PR #27, opened 2017-05-02
Do not retry on 401 failures.

hayderc:
auth failures should not be retried, benefit include:
- perf improvement
- API with limited retries will be set off sooner without this fix
- it's the guidance from W3


hayderc:
I signed it!

PR #28, opened 2017-05-02
Do not retry on 401 failures.

hayderc:
auth failures should not be retried, benefit include:

perf improvement
API with limited retries will be set off sooner without this fix
it's the guidance from W3

### DIFF
--- a/src/main/java/com/android/volley/toolbox/BasicNetwork.java
+++ b/src/main/java/com/android/volley/toolbox/BasicNetwork.java
@@ -157,8 +157,10 @@ public class BasicNetwork implements Network {
                             responseHeaders, false, SystemClock.elapsedRealtime() - requestStart);
                     if (statusCode == HttpStatus.SC_UNAUTHORIZED ||
                             statusCode == HttpStatus.SC_FORBIDDEN) {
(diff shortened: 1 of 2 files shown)


jpd236:
Thanks for sending this patch. However, I don't think it should be merged, nor necessary to accomplish what you'd like. In the future, you may want to send a note to volley-users@googlegroups.com or at least open an issue to save yourself some time up front :)

Changing the default behavior is always risky because other teams using the library may be relying on retries in this manner. So in general, I'd ask that any change like this not change default behavior, but allow apps which are interested in the change to request it explicitly.

However, Volley already provides the necessary hook to make this work - the RetryPolicy. If a particular app doesn't want to retry on 401 failures, use a RetryPolicy which inspects the VolleyError passed to retry() and rethrow it in this case. (You could look whether the VolleyError is an instance of AuthFailureError, or check the network response via VolleyError#networkResponse).

There's actually a very good reason for this which is compatible with the spec you highlighted - while you perhaps shouldn't retry the request without changing the input parameters, a reasonable retry policy might invalidate their auth token and fetch a new one before returning from retry(), which would make it more likely to succeed.

PR #35, opened 2017-05-23
NetworkImageView bug when used with RecyclerView

guelo:
I came across a bug where images would sometimes not show while scrolling a NetworkImageView inside a RecyclerView.  I tracked the bug down to this line with the bitmap getting nulled out in `onDetachedFromWindow`. The problem is that `setImageBitmap(null);` depends on `onLayout` getting triggered after `onAttachedToWindow`, but sometimes while scrolling the RecyclerView ends up calling onAttached -> onDetached -> onAttached but onLayout doesn't get retriggered so the bitmap doesn't get set. Setting the bitmap to null is a memory optimization that is better left up to RecyclerView. In non-RecyclerView situations this optimization might be useful but you can't guarantee that onAttached will always trigger an onLayout pass.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/NetworkImageView.java
+++ b/src/main/java/com/android/volley/toolbox/NetworkImageView.java
@@ -204,7 +204,7 @@ public class NetworkImageView extends ImageView {
             // If the view was bound to an image request, cancel it and clear
             // out the image from the view.
             mImageContainer.cancelRequest();
-            setImageBitmap(null);
+
             // also clear out the container so we can reload the image if necessary.
             mImageContainer = null;
         }


guelo:
I signed it!

jpd236:
We have an app internally which has been using NetworkImageViews in a RecyclerView without any problems. So I'm not convinced this isn't a bug in the calling code. Do you have a small example app which exhibits the problem?

arnaud-web:
Hi,

I reproduce the problem. 

Here is an example: 

https://github.com/arnaud-web/NetworkImageViewPOC

jpd236:
@arnaud-web Do you have more details about what to do with this sample app to see the problem?

arnaud-web:
Hi @jpd236 

You can reproduce the bug by scrolling several times left to right and right to left. 

Take a look at the screenshot image.

Issue #1, opened 2017-01-25
Bring up build infrastructure


jpd236:
Includes the following components:

- Clean up the makefiles (perhaps just use Gradle for everything instead of Maven)
- Continuous integration builds/tests for commits and pull requests
- Other useful things to check in CI (e.g. error-prone, google-java-format)
- Artifact deployment (snapshots on commit, official releases on tags)

jpd236:
Forked off separate bugs for google-java-format and error-prone; these don't need to block the next release.

The only remaining item we need for the next release is ensuring we can promote snapshot builds from OJO to Bintray, including GPG signing, which I'm in the process of investigating.

jpd236:
Unfortunately I'm a bit blocked on GPG signing as I haven't gotten any answers at https://stackoverflow.com/questions/46480065/gpg-signing-snapshots-on-build-promotion-to-bintray around how to GPG sign SNAPSHOT artifacts and/or specify our GPG passphrase when promoting a SNAPSHOT to Bintray. And I really don't want to be publishing an official Volley release from my workstation directly. Currently I'm trying to poke around to see who can answer these questions.

jpd236:
Per JFrog, OJO doesn't support GPG signing for SNAPSHOT artifacts or signing at promotion time via snapshotsToBintray. However, we should be able to sign them manually after the fact. There'll be a brief window where they are not signed, but if all goes well, not much of one.

Will take a shot at this soon.

PR #38, opened 2017-06-01
Cleaned up some duplication

SamuelYvon:
The legacy method was doing the exact same thing as `addIfBodyExists`. Now merged the two and added an override to support the different source of byte stream. Also formatted the file.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/HurlStack.java
+++ b/src/main/java/com/android/volley/toolbox/HurlStack.java
@@ -77,7 +77,7 @@ public class HurlStack implements HttpStack {
     }
 
     /**
-     * @param urlRewriter Rewriter to use for request URLs
+     * @param urlRewriter      Rewriter to use for request URLs
      * @param sslSocketFactory SSL factory to use for HTTPS connections
      */
     public HurlStack(UrlRewriter urlRewriter, SSLSocketFactory sslSocketFactory) {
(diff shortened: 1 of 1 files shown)


SamuelYvon:
I signed it!

src/main/java/com/android/volley/toolbox/HurlStack.java
```diff
@@ -77,7 +77,7 @@ public HurlStack(UrlRewriter urlRewriter) {
     }
 
     /**
-     * @param urlRewriter Rewriter to use for request URLs
+     * @param urlRewriter      Rewriter to use for request URLs
```
jpd236: I'd prefer if you revert the formatting changes for now. This format doesn't match what I think we'd be likely to use in the future (google-java-format), and I'd rather handle that in a more coordinated matter.

src/main/java/com/android/volley/toolbox/HurlStack.java
```diff
@@ -259,6 +254,16 @@ private static void addBodyIfExists(HttpURLConnection connection, Request<?> req
             throws IOException, AuthFailureError {
         byte[] body = request.getBody();
         if (body != null) {
+            addBodyIfExists(connection, request, body);
+        }
+    }
+
+    private static void addBodyIfExists(HttpURLConnection connection, Request<?> request, byte[] body)
+            throws IOException, AuthFailureError {
+        if (body != null) {
```
jpd236: both the callers already check that body != null, so either remove this check here and rename this method "addBody" (my preference), or remove the checks above.

src/main/java/com/android/volley/toolbox/HurlStack.java
SamuelYvon: Makes a lot of sense! I was unaware of the google java format

src/main/java/com/android/volley/toolbox/HurlStack.java
SamuelYvon: Agreed, I will commit the changes when I'm done work!

jpd236:
Thanks for the cleanup. I tweaked your change a bit to revert some extraneous formatting changes and fix the indenting on the new function, but otherwise it looks good.

SamuelYvon:
Sorry for the leftover formatting, I noticed it this morning but I did not wanted to fix it until I was done work. FYI the google java format formatter for intellij/android studio keeps that formatting. I don't know if this is a bug with the formatter or it was just not implemented yet, thought I would let you know.

jpd236:
We haven't yet done a pass of the codebase under google-java-format or made it part of the contribution process. That's being tracked in #1. Until then, I'd just say to try and keep things as consistent with what's there as possible while minimizing extraneous changes. Thanks again.

Issue #21, opened 2017-04-20
Support for multiple identical headers


jpd236:
Previous discussion: https://groups.google.com/forum/#!topic/volley-users/rNTlV-LORzY

The main use case here seems to be server responses with multiple Set-Cookie headers. It looks like a valid feature request, but the concern is that this would require breaking the public API.

mousumikhatua:
I am facing the similar issues.
I get multiple cookie sets with Set-Cookie headers.
In HurlStack.java line 122 - it's only considering the first one.
Need an API update so that I can get all the cookie values.

gonzadocarmo:
So, I'm seeing 2 ways here:
- breaking the public API --> then v2.0.0.
- not breaking the public API --> extending/adapting it so that it does not break the current API. Something equals or similar to @georgiecasey solution (https://github.com/georgiecasey/android-volley-duplicateheadersfix)

jpd236:
I looked a little more at this while [working on cleaning up Volley's Apache HTTP dependency](https://github.com/google/volley/issues/2#issuecomment-296517605), and I think it can be done without breaking the current API.

The new BaseHttpStack interface introduced by the above fix obtains a Map<String, List<String>> of response headers instead of a Map<String, String> from the HTTP stack, which is the first step. From there, we can safely add a new Map<String, List<String> field to NetworkResponse and deprecate the old one while still populating it for backwards compatibility with the first header for a particular key. We'd need to make a similar change to Cache.Entry and read in the duplicate keys in DiskBasedCache (which we can thankfully do without changing the on-disk format, since they're just stored as a flat list; the duplicate entries are dropped on read).

So all told, I think we should be able to fix this without breaking public API, which is good. If someone would like to send a PR for this, I'd recommend waiting until #2 is resolved and then following the above path. Otherwise I will try to get to this at some point soon.

shamolideguzhou:
.

jpd236:
https://github.com/google/volley/pull/91 out for review to implement the approach described above.

jpd236:
This should now be resolved in master and will make it to a 1.0.1-SNAPSHOT build soon.

[...]

PR #40, opened 2017-06-03
Support-multiple-identical-headers

miriban:
I figured out that volley only gets the first value of the identical headers in an HTTP response. So the solution I presented is to concatenate the values of the header and separate them with a symbol.
This symbol is **"\0"** and stored in **Volley.IDENTICAL_HEADER_SEPARATOR**

Developer can easily get the values of the identicale header by splitting it using volley separator. 
`String cookies[] = response.get("Set-Cookie").split(Volley.IDENTICAL_HEADER_SEPARATOR);`

### DIFF
--- a/src/main/java/com/android/volley/toolbox/HurlStack.java
+++ b/src/main/java/com/android/volley/toolbox/HurlStack.java
@@ -16,6 +16,8 @@
 
 package com.android.volley.toolbox;
 
+import android.text.TextUtils;
+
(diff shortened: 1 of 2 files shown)


miriban:
I signed it!

jpd236:
This approach wouldn't be backwards-compatible and could break existing clients that don't expect it.

Please see https://github.com/google/volley/issues/21 for a deeper discussion of how this should be fixed.

PR #52, opened 2017-07-07
DiskBasedCache fix with unit tests

joebowbeer:
Fixes Issue #12 without changing cache format

Changes:

1. DiskBasedCache fixed to avoid NegativeArraySizeException and OutOfMemoryError thrown from streamToBytes. Enhanced CountingInputStream nested class to maintain bytesRemaining (the file length initially), so that streamToBytes can validate its length parameter before attempting to allocate a new byte array.

2. DiskBasedCacheTest unit tests extended to 100% method coverage and 95% code coverage.
    All tests passing: https://circleci.com/gh/joebowbeer/volley/24


joebowbeer:
See https://github.com/google/volley/pull/13 for prior discussion.

jpd236:
Thanks for the rebase :)

I'll get to this review soon - I want to give it a very careful look. At a high level, though, I feel much better about this approach than what we were looking at before.

rules.gradle
```diff
@@ -14,8 +14,9 @@ android {
 // Check if the android plugin version supports unit testing.
 if (configurations.findByName("testCompile")) {
   dependencies {
-    testCompile "junit:junit:4.10"
-    testCompile "org.mockito:mockito-core:1.9.5"
+    testCompile "junit:junit:4.12"
+    testCompile "org.hamcrest:hamcrest-library:1.3"
+    testCompile "org.mockito:mockito-core:2.2.16"
```
jpd236: Any objections to using 2.2.29 instead? It's the latest from the 2.2 branch (and is available in our internal prebuilt repositories, whereas 2.2.16 isn't, so it'll make our lives a bit easier when we pull in this patch).

Alternatively, I'd certainly be fine with leaving this at 1.9.5 and dealing with upgrading Mockito separately. Do you need specific new functionality from the upgraded deps?

jpd236:
@yilanl - feel free to take a look as well to see if you have any further concerns.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -149,28 +149,25 @@ public synchronized void initialize() {
...
-            } finally {
+                CountingInputStream cis = new CountingInputStream(
+                        new BufferedInputStream(createInputStream(file)), (int) file.length());
                 try {
-                    if (fis != null) {
-                        fis.close();
-                    }
-                } catch (IOException ignored) { }
+                    CacheHeader entry = CacheHeader.readHeader(cis);
+                    entry.size = cis.bytesRemaining();
```
joebowbeer: The previous code was inconsistent. The entry.size was set to file.length() in initialize(), but was set to the data.length in put(), via the two-arg CacheHeader constructor. The new code consistently uses the data length.

src/test/java/com/android/volley/toolbox/DiskBasedCacheTest.java
```diff
@@ -18,46 +18,364 @@
...
+
+    @Test
+    public void testIOExceptionInInitialize() throws IOException {
+        // Cache a few kilobytes
+        cache.put("kilobyte", randomData(1024));
+        cache.put("kilobyte2", randomData(1024));
+        cache.put("kilobyte3", randomData(1024));
+
+        // Create DataInputStream that throws IOException
+        InputStream mockedInputStream = spy(InputStream.class);
```
joebowbeer: I think spy is needed, or at the least it is an expedient way to inject only the failure that I want.
I want the methods other than read() to function.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -110,30 +112,28 @@ public synchronized Entry get(String key) {
...
-        CountingInputStream cis = null;
         try {
-            cis = new CountingInputStream(new BufferedInputStream(new FileInputStream(file)));
-            CacheHeader.readHeader(cis); // eat header
-            byte[] data = streamToBytes(cis, (int) (file.length() - cis.bytesRead));
-            return entry.toCacheEntry(data);
+            CountingInputStream cis = new CountingInputStream(
+                    new BufferedInputStream(createInputStream(file)), (int) file.length());
+            try {
+                CacheHeader found = CacheHeader.readHeader(cis);
```
joebowbeer: done

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -110,30 +112,28 @@ public synchronized Entry get(String key) {
...
+                if (!key.equals(found.key)) {
```
joebowbeer: done

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -110,30 +112,28 @@ public synchronized Entry get(String key) {
...
+                    // File was shared by two keys and now holds data for a different entry!
+                    VolleyLog.d("%s: key=%s, found=%s", file.getAbsolutePath(), key, found.key);
+                    removeEntry(key);
```
joebowbeer: I added a comment to clarify.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -110,30 +112,28 @@ public synchronized Entry get(String key) {
...
+                    return null;
+                }
+                byte[] data = streamToBytes(cis, cis.bytesRemaining());
+                return entry.toCacheEntry(data);
+            } finally {
+                //noinspection ThrowFromFinallyBlock
```
joebowbeer: done

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -149,28 +149,25 @@ public synchronized void initialize() {
...
                 try {
-                    if (fis != null) {
-                        fis.close();
-                    }
-                } catch (IOException ignored) { }
+                    CacheHeader entry = CacheHeader.readHeader(cis);
+                    entry.size = cis.bytesRemaining();
+                    putEntry(entry.key, entry);
+                } finally {
+                    //noinspection ThrowFromFinallyBlock
```
joebowbeer: done

src/test/java/com/android/volley/toolbox/DiskBasedCacheTest.java
```diff
@@ -18,46 +18,364 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.Mockito.atLeastOnce;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
 
+@RunWith(RobolectricTestRunner.class)
+@Config(manifest="src/main/AndroidManifest.xml", sdk=21)
```
joebowbeer: I changed to sdk=16, which is the minimum supported by Robolectric.

Note that Robolectric runs sdk=16 by default.

I added the sdk attribute here because I wanted to eliminate a Robolectric warning about not being able to find the manifest, and adding a manifest attribute seems to require an sdk attribute. (The other Robolectric tests still generate this warning.)

AFAIK none of the unit tests under the **test** path are testing sdk=8 compliance. For that, I recommend adding some tests under the **androidTest** path.

src/test/java/com/android/volley/toolbox/DiskBasedCacheTest.java
```diff
@@ -133,4 +453,28 @@ public void publicMethods() throws Exception {
...
+        assertThat(actual.responseHeaders, is(equalTo(expected.responseHeaders)));
+        assertThat(actual.serverDate, is(equalTo(expected.serverDate)));
+        assertThat(actual.softTtl, is(equalTo(expected.softTtl)));
+        assertThat(actual.ttl, is(equalTo(expected.ttl)));
+    }
+
+    private Cache.Entry randomData(int length) {
+        Cache.Entry entry = new Cache.Entry();
+        byte[] data = new byte[length];
+        new Random().nextBytes(data);
```
joebowbeer: I made the random see explicit and added a comment.

rules.gradle
joebowbeer: done

joebowbeer:
Thanks for review! PTAL

joebowbeer:
I think the data size is arguably as good or better. This gives the user a
number they can work with, namely, the size of their own data. This makes
put() easier to implement, and is also more easily tested.

joebowbeer:
PS - trim is performed before put, based on the size of the data. I think
it's a bug that initialize wasn't using the data size as well.

jpd236:
I'd counter that the actual disk usage seems like what is more important to developers - it's what shows up in the platform settings and is attributed as usage to the app. It seems harder for a developer to understand/control what's going on if we just say "max usage not counting for some overhead that changes depending on the number of entries you have in the cache + how many headers the HTTP server happens to return in the response" when all they're really concerned about is saying "don't use more than this much space on disk".

Could you elaborate on what makes that approach difficult?

I'm also okay deferring handling of this issue to its own separate issue, as it's certainly unrelated to the overall goal of fixing the crashes here and adding initial tests. But I would only feel comfortable doing so if this wouldn't introduce a regression where the cache might suddenly grow larger on disk than it would have with older versions of Volley due to changes in accounting.

joebowbeer:
The put() implementation works exactly as it did before: pruneIfNeeded is called first in order to release enough space for the new entry to be written (data.length), and the new entry size is recorded as its data.length.

There is no file written before prune is called, so there is no easy way to prune based on the file length of the new entry.

Previously, initialize() worked differently from put(). During initialize() the file length was used to record the entry size. So the recorded size of an entry would change based on whether the entry was initialized from a previously written file, or put() in the current session. But initialize does not prune, so the total size of the cache on disk would not have changed: it would have been pruned according to the data size not the file size. After initialize(), however, a subsequent put() might cause some previously cached entries to be removed, because the size of the cache after initialize() will appear larger than it did when those entries were put().

If this is really the behavior you want to replicate, then it is easy enough. One liner.

However, trying to put() and prune() based on the size of an entry that has not yet been written would be more difficult to achieve, and I would view this as an incompatible change.

joebowbeer:
BTW, Travis build failure is not related to my change and appears to be due to a flaky test in  RequestQueueIntegrationTest: add_requestProcessedInCorrectOrder

jpd236:
Forked https://github.com/google/volley/issues/57 to resolve the inconsistency in how we track sizes. In general, I prefer to keep changes as small and scoped to the bug they're trying to fix as possible - it makes it easier to understand where behavior changed / revert broken changes as needed. This way we won't have to revert this CL or dig into it more deeply if someone reports a cache size growing unexpected large, for instance :)

Forked https://github.com/google/volley/issues/56 for the flaky test. I retried the CI run until it passed (took two tries).

This change looks good to me. Since it's non-trivial, let me loop in the other Volley code reviewer and give them a chance to voice any concerns before I merge.

jpd236:
Thanks for your work and patience :)

PR #66, opened 2017-08-01
Check that deserialized string length in DiskBasedCache does not overflow int

joebowbeer:
Changes:

* Strengthens the length validation in streamToBytes before casting long to int. This fails faster when deserializing a bad string length. (Issue #12 )
* Eliminates unvalidated `(int) long` casts, which improves code quality.

The overflow check is from Java 8's Math.toIntExact: `(int) length != length`

Unit test added. Without the added overflow check, the new test fails because the expected IOException is not thrown.


jpd236:
LGTM, thanks!

Issue #60, opened 2017-07-25
The way to finish a thread in volley.


xyhuangjinfu:
In volley, to finish a thread such as CacheDiapatcher and NetworkDispatcher, are inturrupt not enough? Whether need the mQuit variable necessarily.
If write like this, hava any problems? those thread can be interrupted in other scenes?
`    public void quit() {
        interrupt();
    }`

`        while (!isInterrupted()) {
            ...................
            try {
                request = mQueue.take();
            } catch (InterruptedException e) {
                return;
            }
        }`

xesam:
你为啥想要单独停止一个volley线程？？

jpd236:
I agree it seems strange that we use a separate mQuit flag rather than just relying on the interrupt status. I don't see why you'd want to interrupt the threads but continue to have them pull ensuing requests off the queue.

So I could get behind a change which does what you describe and uses the interrupt status directly rather than relying on a separate mQuit flag.

joebowbeer:
I agree that this looks fishy, but is there really a problem in practice? It would be great to have unit tests before making subtle changes.

Concerning the proper way to improve this, consider:

* If quit() is called before the thread is started, that is before the thread is alive, then a subsequent call to isInterrupted() may return false (according to spec).

* It is possible for the interrupt flag to be cleared by some other code executing on this thread, e.g., if an InterruptedException is caught or interrupted() is called without reasserting the interrupt flag.

The above are reasons to maintain a dedicated cancellation flag and to check this flag inside the loop:

```
while (!mQuit) { ... }
```

Speaking of catching InterruptedException without reasserting the interrupt flag, note that standard practice is to reassert the interrupt flag before continuing:

```
} catch (InterruptedException e) {
  interrupt(); // restore interrupt flag
  return;
}
```

Arguably, this may not be useful here, as a thread stops running after returning from run()... OTOH, these are public Thread subclasses so without further analysis one can imagine they may themselves be subclassed or executed in surprising ways.

jpd236:
> If quit() is called before the thread is started, that is before the thread is alive, then a subsequent call to isInterrupted() may return false (according to spec).

Personally this feels like a minor concern because in practice, RequestQueue always starts threads immediately after creating them. You'd have to be doing something very custom with your queue or somehow manually be using CacheDispatcher/NetworkDispatcher outside the scope of a RequestQueue for this to be a potential issue, right?

[...]

PR #69, opened 2017-08-02
Work for the issue #60.

xyhuangjinfu:


jpd236:
Per discussion on #60, we went with a different approach here (https://github.com/google/volley/pull/195).

PR #73, opened 2017-08-14
Fix soft TTL for duplicate requests

uhager:
When a soft TTL cache entry exists, that response should be served immediately rather than having duplicate requests wait for the first request's network response. This is done by moving the list of waiting, duplicate requests from the RequestQueue to the CacheDispatcher.

### DIFF
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -71,7 +71,7 @@ public class CacheDispatcher extends Thread {
         mNetworkQueue = networkQueue;
         mCache = cache;
         mDelivery = delivery;
-	mWaitingRequestHandler = new WaitingRequestHandler(this);
+        mWaitingRequestHandler = new WaitingRequestHandler(this);
     }
 
     /**
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -160,7 +160,8 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
                                 try {
                                     mNetworkQueue.put(request);
                                 } catch (InterruptedException e) {
-                                    // Not much we can do about this.
+                                    // Restore the interrupted status
+                                    Thread.currentThread().interrupt();
                                 }
                             }
                         });
@@ -211,21 +212,23 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
         Queue<Request<?>> waitingRequests;
         synchronized (mWaitingRequests) {
             waitingRequests = mWaitingRequests.remove(cacheKey);
-        }
-        if (waitingRequests != null) {
-            if (VolleyLog.DEBUG) {
-                VolleyLog.v("%d waiting requests for cacheKey=%s; resend to network",
-                    waitingRequests.size(), cacheKey);
-            }
-            Request<?> nextInLine = waitingRequests.remove();
-            synchronized (mWaitingRequests) {
-                mWaitingRequests.put(cacheKey, waitingRequests);
-            }
-            try {
-                mNetworkQueue.put(nextInLine);
-            } catch (InterruptedException iex) {
-                VolleyLog.e("Couldn't add request to queue. %s", iex.toString());
-                quit();
+            if (waitingRequests != null) {
+                if (VolleyLog.DEBUG) {
+                    VolleyLog.v("%d waiting requests for cacheKey=%s; resend to network",
+                        waitingRequests.size(), cacheKey);
+                }
+                Request<?> nextInLine = waitingRequests.remove();
+                if (waitingRequests.size() > 0) {
+                  mWaitingRequests.put(cacheKey, waitingRequests);
+                }
+                try {
+                    mNetworkQueue.put(nextInLine);
+                } catch (InterruptedException iex) {
+                    VolleyLog.e("Couldn't add request to queue. %s", iex.toString());
+                    // Restore the interrupted status
+                    Thread.currentThread().interrupt();
+                    quit();
+                }
             }
         }
     }
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -218,9 +218,7 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
                         waitingRequests.size(), cacheKey);
                 }
                 Request<?> nextInLine = waitingRequests.remove();
-                if (waitingRequests.size() > 0) {
-                  mWaitingRequests.put(cacheKey, waitingRequests);
-                }
+                mWaitingRequests.put(cacheKey, waitingRequests);
                 try {
                     mNetworkQueue.put(nextInLine);
                 } catch (InterruptedException iex) {
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -33,7 +33,7 @@ import java.util.concurrent.BlockingQueue;
  * refresh are enqueued on the specified network queue for processing
  * by a {@link NetworkDispatcher}.
  */
-public class CacheDispatcher extends Thread implements Request.NetworkRequestCompleteListener {
+public class CacheDispatcher extends Thread {
 
     private static final boolean DEBUG = VolleyLog.DEBUG;
 
@@ -52,17 +52,8 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
     /** Used for telling us to die. */
     private volatile boolean mQuit = false;
 
-    /**
-     * Staging area for requests that already have a duplicate request in flight.
-     *
-     * <ul>
-     *     <li>containsKey(cacheKey) indicates that there is a request in flight for the given cache
-     *          key.</li>
-     *     <li>get(cacheKey) returns waiting requests for the given cache key. The in flight request
-     *          is <em>not</em> contained in that list. Is null if no requests are staged.</li>
-     * </ul>
-     */
-    private final Map<String, Queue<Request<?>>> mWaitingRequests = new HashMap<>();
+    /** Manage list of waiting requests and de-duplicate requests with same cache key. */
+    private WaitingRequestHandler mWaitingRequestHandler;
 
     /**
      * Creates a new cache triage dispatcher thread.  You must call {@link #start()}
@@ -80,6 +71,7 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
         mNetworkQueue = networkQueue;
         mCache = cache;
         mDelivery = delivery;
+	mWaitingRequestHandler = new WaitingRequestHandler();
     }
 
     /**
@@ -117,7 +109,7 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
                 if (entry == null) {
                     request.addMarker("cache-miss");
                     // Cache miss; send off to the network dispatcher.
-                    if (!maybeAddToWaitingRequests(request)) {
+                    if (!mWaitingRequestHandler.maybeAddToWaitingRequests(request)) {
                         mNetworkQueue.put(request);
                     }
                     continue;
@@ -127,7 +119,7 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
                 if (entry.isExpired()) {
                     request.addMarker("cache-hit-expired");
                     request.setCacheEntry(entry);
-                    if (!maybeAddToWaitingRequests(request)) {
+                    if (!mWaitingRequestHandler.maybeAddToWaitingRequests(request)) {
                         mNetworkQueue.put(request);
                     }
                     continue;
@@ -151,7 +143,7 @@ public class CacheDispatcher extends Thread implements Request.NetworkRequestCom
                     // Mark the response as intermediate.
                     response.intermediate = true;
 
-                    if (!maybeAddToWaitingRequests(request)) {
+                    if (!mWaitingRequestHandler.maybeAddToWaitingRequests(request)) {
                         // Post the intermediate response back to the user and have
                         // the delivery then forward the request along to the network.
                         mDelivery.postResponse(request, response, new Runnable() {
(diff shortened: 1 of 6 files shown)


src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +179,92 @@ public void run() {
...
+
+    /**
+     * For cacheable requests, if a request for the same cache key is already in flight,
+     * add it to a queue to wait for that in-flight request to finish.
+     * @return whether the request was queued. If false, we should continue issuing the request
+     * over the network. If true, we should put the request on hold to be processed when
+     * the in-flight request finishes.
+     */
+    private boolean maybeAddToWaitingRequests(Request<?> request) {
+        String cacheKey = request.getCacheKey();
```
jjoslin: nit: make this final

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -128,22 +148,27 @@ public void run() {
...
+                    if (!maybeAddToWaitingRequests(request)) {
+                        // Post the intermediate response back to the user and have
+                        // the delivery then forward the request along to the network.
+                        mDelivery.postResponse(request, response, new Runnable() {
+                            @Override
+                            public void run() {
+                                try {
+                                    mNetworkQueue.put(request);
+                                } catch (InterruptedException e) {
+                                    // Not much we can do about this.
```
jjoslin: Don't just swallow the InterruptedException, reset the interrupted status on the thread.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +179,92 @@ public void run() {
...
+                    waitingRequests.size(), cacheKey);
+            }
+            Request<?> nextInLine = waitingRequests.remove();
+            synchronized (mWaitingRequests) {
+                mWaitingRequests.put(cacheKey, waitingRequests);
+            }
+            try {
+                mNetworkQueue.put(nextInLine);
+            } catch (InterruptedException iex) {
+                VolleyLog.e("Couldn't add request to queue. %s", iex.toString());
```
jjoslin: Don't just swallow the InterruptedException, reset the interrupted status on the thread.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +179,92 @@ public void run() {
...
+
+    /** No valid response received from network, release waiting requests. */
+    @Override
+    public void onNoUsableResponseReceived(Request<?> request) {
+        String cacheKey = request.getCacheKey();
+        Queue<Request<?>> waitingRequests;
+        synchronized (mWaitingRequests) {
+            waitingRequests = mWaitingRequests.remove(cacheKey);
+        }
+        if (waitingRequests != null) {
```
jjoslin: This is racy and you should expand the synchronized block above to also include the mWaitingRequests.put() call below. If you don't then the map can be modified by maybeAddToWaitingRequests() before the code hits the synchronized block below and any changes will be clobbered.

src/main/java/com/android/volley/Request.java
```diff
@@ -585,6 +599,36 @@ public void deliverError(VolleyError error) {
...
+        mRequestCompleteListener = requestCompleteListener;
+    }
+
+    /**
+     * Notify NetworkRequestCompleteListener that a valid response has been received
+     * which can be used for other, waiting requests.
+     * @param response received from the network
+     */
+    /* package */ void notifyListenerResponseReceived(Response<?> response) {
+        if (mRequestCompleteListener != null) {
```
jjoslin: Access to mRequestCompleteListener should be synchronized. I believe the code works in its current state because mRequestCompleteListener is never concurrently accessed but if the code is ever rearranged then this could be a problem.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -49,6 +53,18 @@
...
+     * Staging area for requests that already have a duplicate request in flight.
+     *
+     * <ul>
+     *     <li>containsKey(cacheKey) indicates that there is a request in flight for the given cache
+     *          key.</li>
+     *     <li>get(cacheKey) returns waiting requests for the given cache key. The in flight request
+     *          is <em>not</em> contained in that list. Is null if no requests are staged.</li>
+     * </ul>
+     */
+    private final Map<String, Queue<Request<?>>> mWaitingRequests = new HashMap<>();
```
joebowbeer: Can you remove explicit synchronization of waiting requests and replace with ConcurrentHashMap instance? The downside is that you cannot use null as a marker, but that may be a good thing:-)

Is the CHM putIfAbsent method useful?

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +180,92 @@ public void run() {
...
+    private boolean maybeAddToWaitingRequests(Request<?> request) {
+        String cacheKey = request.getCacheKey();
+        synchronized (mWaitingRequests) {
+            // Insert request into stage if there's already a request with the same cache key
+            // in flight.
+            if (mWaitingRequests.containsKey(cacheKey)) {
+                // There is already a request in flight. Queue up.
+                Queue<Request<?>> stagedRequests = mWaitingRequests.get(cacheKey);
+                if (stagedRequests == null) {
+                    stagedRequests = new LinkedList<Request<?>>();
```
joebowbeer: ArrayList is more performant here (and virtually everywhere). In other words, LinkedList is best avoided in Java.

src/main/java/com/android/volley/Request.java
```diff
@@ -585,6 +603,42 @@ public void deliverError(VolleyError error) {
     }
 
     /**
+     * {@link NetworkRequestCompleteListener} that will receive callbacks when the request
+     * returns from the network.
+     */
+    /* package */ void setNetworkRequestCompleteListener(
+            NetworkRequestCompleteListener requestCompleteListener) {
+        synchronized(mLock) {
```
joebowbeer: code style nit:

```
synchronized (mLock) {
```

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +180,92 @@ public void run() {
...
+                mDelivery.postResponse(waiting, response);
+            }
+        }
+    }
+
+    /** No valid response received from network, release waiting requests. */
+    @Override
+    public void onNoUsableResponseReceived(Request<?> request) {
+        String cacheKey = request.getCacheKey();
+        Queue<Request<?>> waitingRequests;
```
joebowbeer: Move this to where it is assigned.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +180,92 @@ public void run() {
...
+        synchronized (mWaitingRequests) {
+            waitingRequests = mWaitingRequests.remove(cacheKey);
+            if (waitingRequests != null) {
+                if (VolleyLog.DEBUG) {
+                    VolleyLog.v("%d waiting requests for cacheKey=%s; resend to network",
+                        waitingRequests.size(), cacheKey);
+                }
+                Request<?> nextInLine = waitingRequests.remove();
```
joebowbeer: How do you know nextInLine is not null? (A BlockingQueue does not accept null elements.)

src/main/java/com/android/volley/CacheDispatcher.java
jpd236: I think we should probably defer discussion of that kind of change to a separate PR / bug if desired. This portion of the code was just moved from one class to another, and to keep the scope of this PR manageable we should try to focus it solely on the bug it is trying to fix rather than introduce lots of other small fixes. (I do not think this would be a trivial change; often ConcurrentHashMap has significantly worse performance than a simple HashMap with a lock around it for low levels of concurrency).

src/main/java/com/android/volley/CacheDispatcher.java
jpd236: This one was also just copied from the old code... but in this case I'm fine with making the change as I do think it's reasonable enough.

jpd236:
Also, @joebowbeer, thanks so much for the review! We appreciate your help :)

src/main/java/com/android/volley/CacheDispatcher.java
joebowbeer: @jpd236 Understood. Every time I see `synchronized` added, it raises my concern about maintainability, correctness and potential deadlock. I doubt ConcurrentHashMap would have any negative impact on the performance of this method at any level of concurrency. It might lead to a design that is easier to understand, though, and avoid questions about correctness. I felt obliged to raise this issue but I won't push back.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -128,22 +148,28 @@ public void run() {
...
-                            } catch (InterruptedException e) {
-                                // Not much we can do about this.
+                    if (!maybeAddToWaitingRequests(request)) {
+                        // Post the intermediate response back to the user and have
+                        // the delivery then forward the request along to the network.
+                        mDelivery.postResponse(request, response, new Runnable() {
+                            @Override
+                            public void run() {
+                                try {
+                                    mNetworkQueue.put(request);
```
joebowbeer: Just wondering: What thread executes this blocking call? (I realize that this is just moving code around.)

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -29,7 +33,7 @@
  * refresh are enqueued on the specified network queue for processing
  * by a {@link NetworkDispatcher}.
  */
-public class CacheDispatcher extends Thread {
+public class CacheDispatcher extends Thread implements Request.NetworkRequestCompleteListener {
```
joebowbeer: Can you push this listener responsibility into a delegate? Best practice is generally to limit scope/capability as much as possible... I bring this up because it is confusing to me to see new methods added to this Thread class that do not execute on this Thread.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +180,92 @@ public void run() {
...
+                mWaitingRequests.put(cacheKey, stagedRequests);
+                if (VolleyLog.DEBUG) {
+                    VolleyLog.d("Request for cacheKey=%s is in flight, putting on hold.", cacheKey);
+                }
+                return true;
+            } else {
+                // Insert 'null' queue for this cacheKey, indicating there is now a request in
+                // flight.
+                mWaitingRequests.put(cacheKey, null);
+                request.setNetworkRequestCompleteListener(this);
```
joebowbeer: Make `this` a dedicated listener instance? (See comment above.)

joebowbeer:
@jpd236 regarding code coverage, in Android Studio right-click on `com.android.volley (test)` and select *Run Tests with Coverage*. The coverage of CacheDispatcher prior to this PR was nearly 100%. It would be cool to add this to the Travis build.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +180,92 @@ public void run() {
...
+                }
+                Request<?> nextInLine = waitingRequests.remove();
+                mWaitingRequests.put(cacheKey, waitingRequests);
+                try {
+                    mNetworkQueue.put(nextInLine);
+                } catch (InterruptedException iex) {
+                    VolleyLog.e("Couldn't add request to queue. %s", iex.toString());
+                    // Restore the interrupted status
+                    Thread.currentThread().interrupt();
+                    quit();
```
joebowbeer: This double interrupt is confusing. Worth added a comment explaining which thread is executing this listener method, and pointing out that Thread.currentThread().interrupt() is reasserting the interrupt on the calling thread, while the call to interrupt() inside quit() is interrupting the CacheDispatcher thread?

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Wouldn't we still have to add a synchronized block in onNoUsableResponseReceived anyway?

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Done.

src/main/java/com/android/volley/CacheDispatcher.java
uhager: I'm not quite sure which call you are referring to, but the CacheDispatcher runs on its own thread. In the default implementation, the Runnable is executed in ExecutorDelivery, which runs it on a thread that is set in ExecutorDelivery's instantiation by RequestQueue, where the default is the main looper. The network dispatcher then runs again on its own thread when it grabs the request from the NetworkQueue.

src/main/java/com/android/volley/CacheDispatcher.java
joebowbeer: @uhager synchronized can always be avoided :-) java.util.concurrent was written without a single synchronized. My advice to use CHM may be a bum lead, but the motivation is to make the synchronization more local and explicit. CHM may not be the best way to do that, but it feels like it may be. Seeing `synchronized` added to the code without explanation of which threads are involved, etc., worries me. Would moving the listener methods to a nested delegate, as I suggest,  also allow the new synchronization to be more localized?

src/main/java/com/android/volley/CacheDispatcher.java
joebowbeer: @uhager I was referring to the `mNetworkQueue.put` which is a blocking call (hence the InterruptedException).

src/main/java/com/android/volley/CacheDispatcher.java
uhager: This is a Queue though. Should I make the Queue a List instead?

joebowbeer: @uhager Ah. Use `ArrayDeque`:

```
This class is likely to be faster than Stack when used as a stack,
and faster than LinkedList when used as a queue.
```

Null items are prohibited. I hope that's not a problem.

jpd236: Volley's minSdkVersion is still 8, and ArrayDeque was added in API 9. Definitely a reasonable change to consider bumping minSdkVersion and making this change but it'd be out of the scope of the CL. So probably best to leave as is for now.

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Added a check.

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Do you mean as a nested class in CacheDispatcher?

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Ah, that one then runs on the main thread (when using the default RequestQueue with the default ExecutorDelivery).

src/main/java/com/android/volley/CacheDispatcher.java
joebowbeer: @uhager Yes. Move the listener methods, the mWaitingRequests instance and the `maybeAdd` helper method to a nested class. Then you can just synchronize the methods of that nested class, and it will be very clear what state is being maintained.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -48,6 +52,9 @@
     /** Used for telling us to die. */
     private volatile boolean mQuit = false;
 
+    /** Manage list of waiting requests and de-duplicate requests with same cache key. */
+    private WaitingRequestHandler mWaitingRequestHandler;
```
joebowbeer: make this final

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -64,6 +71,7 @@ public CacheDispatcher(
         mNetworkQueue = networkQueue;
         mCache = cache;
         mDelivery = delivery;
+	mWaitingRequestHandler = new WaitingRequestHandler();
```
joebowbeer: Declare static nested so the interactions with CacheDispatcher are explicit: `WaitingRequestHandler(this)`

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +172,104 @@ public void run() {
             }
         }
     }
+
+    private class WaitingRequestHandler implements Request.NetworkRequestCompleteListener {
```
joebowbeer: `private static class` - to make the interactions with CacheDispatcher explicit.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +172,104 @@ public void run() {
...
+
+	/** Request received a valid response that can be used by other waiting requests. */
+	@Override
+	public synchronized void onResponseReceived(Request<?> request, Response<?> response) {
+	    if (response.cacheEntry == null || response.cacheEntry.isExpired()) {
+		onNoUsableResponseReceived(request);
+		return;
+	    }
+	    String cacheKey = request.getCacheKey();
+	    Queue<Request<?>> waitingRequests;
```
joebowbeer: declaration and assignment on single line.

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Thanks for all the great comments! I'm new to Java, so I'm learning a lot here

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -64,6 +71,7 @@ public CacheDispatcher(
         mNetworkQueue = networkQueue;
         mCache = cache;
         mDelivery = delivery;
+	mWaitingRequestHandler = new WaitingRequestHandler(this);
```
jjoslin: nit: the indentation here seems to be off by 1 space.

uhager: Done.

jjoslin:
@joebowbeer I think that's a reasonable suggestion but I'd prefer to see it done as a separate pull request.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -48,6 +52,9 @@
     /** Used for telling us to die. */
     private volatile boolean mQuit = false;
 
+    /** Manage list of waiting requests and de-duplicate requests with same cache key. */
+    private final WaitingRequestHandler mWaitingRequestHandler;
```
jpd236: nit: Handler is a loaded term on Android (https://developer.android.com/reference/android/os/Handler.html), so for clarity's sake it would be good to avoid using it if we're not actually extending that class.

Perhaps "WaitingRequestManager" or "WaitingRequestDispatcher"?

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +172,109 @@ public void run() {
...
+
+	private final CacheDispatcher mCacheDispatcher;
+	
+	WaitingRequestHandler(CacheDispatcher cacheDispatcher) {
+	    mCacheDispatcher = cacheDispatcher;
+	}
+	
+	/** Request received a valid response that can be used by other waiting requests. */
+	@Override
+	public synchronized void onResponseReceived(Request<?> request, Response<?> response) {
```
jpd236: A key change here (since the last time I've looked at this) is that you are now holding a lock (via synchronized on the method) for all of the calls to postResponse. Previously it was not held during that time. I think this is probably reasonable given that a sane implementation will simply dispatch the response without actually waiting for the response to be handled (and that's the default behavior of ExecutorDelivery) but it's perhaps worth noting that a different delivery (i.e. ImmediateResponseDelivery, though this is only in the test code) would cause this lock to be held during response callback handling.

To summarize, no change requested here, but it'd be good to think through this angle one more time to make sure that there isn't a major increase in contention by holding a lock for each of these methods' entire bodies.

src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -154,4 +172,109 @@ public void run() {
...
+		Request<?> nextInLine = waitingRequests.remove();
+		if (nextInLine == null) {
+		    return;
+		}
+		mWaitingRequests.put(cacheKey, waitingRequests);
+		try {
+		    mCacheDispatcher.mNetworkQueue.put(nextInLine);
+		} catch (InterruptedException iex) {
+		    VolleyLog.e("Couldn't add request to queue. %s", iex.toString());
+		    // Restore the interrupted status of the calling thread (i.e. NetworkDIspatcher)
```
jpd236: nit: NetworkDispatcher

src/main/java/com/android/volley/CacheDispatcher.java
joebowbeer: I was going to comment on this as well: "Strive for *open calls* to avoid deadlock" being the best practice. In other words, don't hold a lock when calling other code.

In this case, the lock can be shrunk quite a bit:

[Hey. There are HARD tabs in this file. Is that cool?]

```
    Queue<Request<?>> waitingRequests;
    synchronized (this) {
        waitingRequests = mWaitingRequests.remove(cacheKey);
    }
    if (waitingRequests != null) {
        if (VolleyLog.DEBUG) {
            VolleyLog.v("Releasing %d waiting requests for cacheKey=%s.",
                    waitingRequests.size(), cacheKey);
        }
        // Process all queued up requests.
        for (Request<?> waiting : waitingRequests) {
            mCacheDispatcher.mDelivery.postResponse(waiting, response);
        }
    }
```

jpd236: Yeah, I think that's how it was in the initial PR, so we should probably move back to that unless there is a thread-safety issue. Just need to treat real carefully there :)

And for tabs, I'm not as used to GitHub's review tool so I haven't caught whitespace issues like that, but yes, we should be using spaces and not tabs. https://github.com/google/volley/issues/1 mentions integrating google-java-format into the workflow here which would resolve that more automatically FYI.

uhager: I had that initially, but made it all synchronized for the inner class to avoid synchronized blocks. I thought that was part of the point of the nested class to have clearer synchronization by synchronizing the methods? But I can revert that. 
Apologies for the tabs, I was working on that in emacs at home, I'll have to check my configurations.

src/main/java/com/android/volley/CacheDispatcher.java
uhager: Went with Manager.

src/main/java/com/android/volley/CacheDispatcher.java
joebowbeer: @uhager sorry for the added confusion. The inner class separates concerns nicely, but the need for open calls adds a small complication. Thanks for addressing.

PR #75, opened 2017-08-21
Deprecate Volley's use of Apache HTTP.

### LINKED ISSUE #2: Make the dependency on Apache HTTP optional
Volley currently depends on Apache HTTP. We can mostly remove the dependency except that HttpStack#performRequest returns an org.apache.http.HttpResponse, which means it'd be an API-breaking change to remove.

It might be worth biting the bullet and just doing a major version bump that breaks the API. There are other known issues that we could potentially address alongside this. However, one avenue worth exploring would be to see if we could treat Apache HTTP as a "provided" dependency - that is, one which is available at compile time, but not runtime - and then deprecate the old performRequest and add a new one that is implemented by all built-in clients (except HttpClientStack, which should be deprecated as well).

I *think* the result would compile and let apps build against Volley without pulling in the Apache HTTP dependency, but existing users could continue to use the Apache stack as long as they pulled in the dependency at runtime (and it should be there on an actual Android device).

Apart from the ugliness, my main concern is how proguard would respond to the missing reference - we could add a consumer proguard rule to not warn on org.apache stuff though that might be a bit overbearing.

jpd236:
This change makes it feasible for Volley clients to remove their dependency on the legacy Apache HTTP library; if HurlStack is used, or if their existing HttpStack implementation is refactored atop the new BaseHttpStack (which should be trivial), then Volley will never depend on classes in org.apache at runtime. However, legacy clients built atop HttpStack (including HttpClientStack) should continue to work without modifications for now.

Fixes #2


src/main/java/com/android/volley/toolbox/BaseHttpStack.java
```diff
@@ -0,0 +1,95 @@
...
+                protocolVersion, response.getStatusCode(), "" /* reasonPhrase */);
+        BasicHttpResponse apacheResponse = new BasicHttpResponse(statusLine);
+
+        List<Header> headers = new ArrayList<>();
+        for (Map.Entry<String, List<String>> entry : response.getHeaders().entrySet()) {
+            for (String value : entry.getValue()) {
+                headers.add(new BasicHeader(entry.getKey(), value));
+            }
+        }
+        apacheResponse.setHeaders(headers.toArray(new Header[0]));
```
jjoslin: If you allocate the correct size up front, headers.toArray(new Header[headers.size()]), then you can avoid an unnecessary allocation of an array.

src/main/java/com/android/volley/toolbox/HttpHeaderParser.java
```diff
@@ -126,13 +133,25 @@
     public static long parseDateAsEpoch(String dateStr) {
         try {
             // Parse date in RFC1123 format if this header contains one
-            return DateUtils.parseDate(dateStr).getTime();
-        } catch (DateParseException e) {
+            return newRfc1123Formatter().parse(dateStr).getTime();
+        } catch (ParseException e) {
             // Date in invalid format, fallback to 0
```
jjoslin: I realize this is out of scope for this review but it seems like this exception should be logged.

src/main/java/com/android/volley/NetworkResponse.java
```diff
@@ -48,11 +47,11 @@ public NetworkResponse(int statusCode, byte[] data, Map<String, String> headers,
     }
 
     public NetworkResponse(byte[] data) {
-        this(HttpStatus.SC_OK, data, Collections.<String, String>emptyMap(), false, 0);
+        this(HttpURLConnection.HTTP_OK, data, Collections.<String, String>emptyMap(), false, 0);
```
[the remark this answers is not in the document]
jpd236: I always thought this was against the style guide, but it doesn't look like it. Well, I guess I personally prefer having the verbosity since I feel like imports are generally not noticed when looking at a file and the source of these constants is kind of relevant, though I wouldn't push back if a static import were used elsewhere. Going to just leave this as is to avoid churn.

src/main/java/com/android/volley/toolbox/BaseHttpStack.java
jpd236: Done.

src/main/java/com/android/volley/toolbox/HttpHeaderParser.java
jpd236: Agreed; added log.

Issue #76, opened 2017-08-24
Can't read "contentType" from cache


dingbuoyi:
I have met a issue , when I try to get the "contentType" from cache, it always return null.
Then I checked the source code of "DiskBasedCache" and "BasicNetwork".

It turns out that in "DiskBasedCache" , the volley use "HaspMap" to save the http headers and in 
"BasicNetwork" it uses "new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER)" which
is not case sensitive.

So if the server's response header, uses "content-type" rather than "Content-Type", it can work properly with a response from network but can't get contentType from the cache.

And because of this issue , you can't parse the charset from cache  , and it will use the default charset which is "ISO_8859_1" , and if it doesn't match the actually charset which you should get from "content-type" , then you can't get the right response content with the right charset from cache.

joebowbeer:
I think I see what you're referring to.

HttpHeaderParser is *expecting* to see a normalized header map: [HttpHeaderParser#L145](https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/HttpHeaderParser.java#L145)

and BasicNetwork converts its headers to a string-insensitive map before they are cached:  [BasicNetwork#L101](https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/BasicNetwork.java#L101)

BUT if a cached entry is returned, its header map is parsed, and its map will not perform case-insensitive header comparisons: [CacheDispatcher#L119](https://github.com/google/volley/blob/master/src/main/java/com/android/volley/CacheDispatcher.java#L119)

It should be easy to write a failing unit test.

dingbuoyi:
@joebowbeer 

Thank you for reply.  You completely get my point.
Should I write the unit test or you will?

joebowbeer:
@dingbuoyi feel free to write the unit test and/or PR

Looking at this a little closer, I think the fix needs to be at the sites where the NetworkResponse is created. The Cache interface doesn't spec this contract currently.

In particular, CacheDispatcher should convert the headers before passing them to the NetworkResponse constructor: [CacheDispatcher#L118](https://github.com/google/volley/blob/master/src/main/java/com/android/volley/CacheDispatcher.java#L118)

```
    Map<String, String> responseHeaders =
            new TreeMap<>(String.CASE_INSENSITIVE_ORDER).putAll(entry.responseHeaders);
    Response<?> response = request.parseNetworkResponse(
[...]

PR #79, opened 2017-08-29
all the headers pass to NetworkResponse should be case-insensitive

dingbuoyi:
https://github.com/google/volley/issues/76 Can't get "contentType" correctly from cache when using "DiskBasedCache".

It's not good to force request all the cache implementation using an case-insensitive map to cache the headers.

So we need to convert them before we pass them to the "NetworkResponse".


dingbuoyi:
I signed it!

dingbuoyi:
I signed it!

dingbuoyi:
I signed it!

dingbuoyi:
I signed it!

PR #86, opened 2017-09-10
Create StringResponseModel, available for modify html elements 

gichang-yang:
current Volley's StringRequest is Just receive raw html body. 
However, Some websites are not fit in 'Mobile'. 

screenshot_2017-09-10-19-54-18

look at this screen shot. as original web site is not fit in mobile, img element is too big to use in mobile.

So, we need modify & adjusting method in 'ResponsedString' Class. 
also, other modify or adjusting methods could be add in this class.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/ParsedString.java
+++ b/src/main/java/com/android/volley/toolbox/ParsedString.java
@@ -1,0 +1,60 @@
+package com.android.volley.toolbox;
+
+
+import java.io.UnsupportedEncodingException;
+
+/**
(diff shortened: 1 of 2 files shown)


jpd236:
Thanks for sending this. However, this kind of site-specific processing is out of scope of what Volley is seeking to accomplish; parsing of responses beyond trivial unmarshaling (like StringRequest or JsonRequest) is the responsibility of the client app. So we wouldn't accept a change like this into the core Volley repository.

As an app, however, you should feel free to override StringRequest or any other Request as desired and alter it to fit your app's needs.

PR #89, opened 2017-09-12
For waiting requests, use ArrayList instead of LinkedList

uhager:
As discussed previously, ArrayList is preferred over LinkedList.

### DIFF
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -18,10 +18,10 @@ package com.android.volley;
 
 import android.os.Process;
 
+import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
-import java.util.Queue;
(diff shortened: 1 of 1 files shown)


src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -221,13 +221,13 @@ public void onResponseReceived(Request<?> request, Response<?> response) {
...
-            if (waitingRequests != null) {
+            List<Request<?>> waitingRequests = mWaitingRequests.remove(cacheKey);
+            if (waitingRequests != null && !waitingRequests.isEmpty()) {
                 if (VolleyLog.DEBUG) {
                     VolleyLog.v("%d waiting requests for cacheKey=%s; resend to network",
                             waitingRequests.size(), cacheKey);
                 }
-                Request<?> nextInLine = waitingRequests.remove();
+                Request<?> nextInLine = waitingRequests.remove(0);
                     if (nextInLine == null) {
```
joebowbeer: This check isn't needed anymore and should be removed

jpd236:
Thanks for following up on this!

PR #91, opened 2017-09-14
Improve Volley header handling.

### LINKED ISSUE #21: Support for multiple identical headers

### LINKED ISSUE #76: Can't read "contentType" from cache

jpd236:
-Add support for multiple headers with the same key. This is permitted
by HTTP standards - for example, a server may include multiple
Set-Cookie headers in a response. However, Volley used Map<String,
String> throughout its stack which made it impossible to see all but
the last such header. We now expose a flat Header list in
NetworkResponse and Cache.Entry for clients who wish to use it. The
old Map is still populated for backwards compatibility and
convenience.

-Fix inconsistent case-insensitivity in Header maps. Previously, the
header map would be case-insensitive for network requests but
case-sensitive for cache entry reads. We now ensure that the map is
always case-insensitive (for clients using it and not the flat list)
by creating it in NetworkResponse's constructor. Redundant TreeMap
creations have been removed.

-Avoid mutating the cache entry's response header map directly.

This change should be compile-time-compatible with existing Volley 1.0
users. There is a signature change to HttpResponse which has not been
part of an official release yet. Custom implementations of Cache may
wish to populate the new allHeaders field in Cache.Entry, though this
is optional.

Fixes #21
Fixes #76

### DIFF
--- a/src/main/java/com/android/volley/NetworkResponse.java
+++ b/src/main/java/com/android/volley/NetworkResponse.java
@@ -19,7 +19,6 @@ package com.android.volley;
 import java.net.HttpURLConnection;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
@@ -101,7 +100,7 @@ public class NetworkResponse {
         this.statusCode = statusCode;
         this.data = data;
         this.headers = headers;
-        this.allHeaders = allHeaders;
+        this.allHeaders = Collections.unmodifiableList(allHeaders);
         this.notModified = notModified;
         this.networkTimeMs = networkTimeMs;
     }
@@ -113,9 +112,9 @@ public class NetworkResponse {
     public final byte[] data;
 
     /**
-     * Immutable response headers.
+     * Response headers.
      *
-     * <p>This map is case-insensitive.
+     * <p>This map is case-insensitive. It should not be mutated directly.
      *
      * <p>Note that if the server returns two headers with the same (case-insensitive) name, this
      * map will only contain the last one. Use {@link #allHeaders} to inspect all headers returned
@@ -123,7 +122,7 @@ public class NetworkResponse {
      */
     public final Map<String, String> headers;
 
-    /** All response headers. */
+    /** All response headers. Must not be mutated directly. */
     public final List<Header> allHeaders;
 
     /** True if the server returned a 304 (Not Modified). */
--- a/src/main/java/com/android/volley/NetworkResponse.java
+++ b/src/main/java/com/android/volley/NetworkResponse.java
@@ -17,13 +17,18 @@
 package com.android.volley;
 
 import java.net.HttpURLConnection;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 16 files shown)


src/main/java/com/android/volley/NetworkResponse.java
```diff
@@ -17,56 +17,136 @@
...
+     *
+     * <p>This map is case-insensitive.
+     *
+     * <p>Note that if the server returns two headers with the same (case-insensitive) name, this
+     * map will only contain the last one. Use {@link #allHeaders} to inspect all headers returned
+     * by the server.
+     */
     public final Map<String, String> headers;
 
+    /** All response headers. */
```
joebowbeer: Why is the first header Map documented as immutable and not this one as well? (And are either of these really immutable?)

jpd236: Good question.

I've added a new commit to try and resolve this lack of clarity. The commit description there has full details, but in short - I've made all of the new List<Header> fields strictly immutable and documented them as such (with the exception of Cache.Entry, which is unenforceable because the field itself is mutable), and I've documented the legacy Map<String, String> fields as "should not be mutated" while not enforcing this to avoid breaking legacy apps.

src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -313,16 +313,43 @@ protected void logError(String what, String url, long start) {
...
+        Set<String> headerNamesFromNetworkResponse =
+                new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        for (Header header : responseHeaders) {
+            headerNamesFromNetworkResponse.add(header.getName());
+        }
+
+        // Second, add headers from the cache entry to the network response as long as
+        // they didn't appear in the network response, which should take precedence.
+        List<Header> combinedHeaders = new ArrayList<>(responseHeaders);
+        if (entry.allResponseHeaders != null) {
```
jjoslin: I think if you also check for emptiness here you'll save an iterator allocation (when the collection is empty).

jpd236: Thanks, done. (Though in practice I think it'd be pretty surprising for an HTTP response to not include any response headers).

jjoslin: Agreed.

PR #93, opened 2017-09-18
Force "sudo: false" so we use container builds.

jpd236:


jpd236:
Doesn't seem to make a difference to startup times.

PR #96, opened 2017-09-20
Fix NPEs/compile errors introduced by header changes.

### LINKED ISSUE #95: NetworkDispatcher.run: Unhandled exception java.lang.NullPointerException
I have been using Volley for a while now, but just updated the library today and starting running into some issues. My REST calls were failing with the error 

    [285580] NetworkDispatcher.run: Unhandled exception java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference                                              
        at java.lang.String$CaseInsensitiveComparator.compare(String.java:1104)                                                              
        at java.lang.String$CaseInsensitiveComparator.compare(String.java:1103)
        at java.util.TreeMap.put(TreeMap.java:554)
        at com.android.volley.NetworkResponse.toHeaderMap(NetworkResponse.java:145)
        at com.android.volley.NetworkResponse.<init>(NetworkResponse.java:0)
        at com.android.volley.toolbox.BasicNetwork.performRequest(BasicNetwork.java:169)
        at com.android.volley.NetworkDispatcher.run(NetworkDispatcher.java:113)

From my searching I found it was issues with the headers. I did look at my headers and I could not find issues with them. I started searching more and lead me to `NetworkResponse.java` file in the Volley library. In the function `toHeaderMap`, there is a header with a null name:

    Header[name=null,value=HTTP/1.1 201 Created]

Which does follow the standard request header

    HTTP/1.1 200 OK
    Cache-Control: private
    Content-Type: text/plain; charset=utf-8
    Content-Encoding: gzip
    Vary: Accept-Encoding 
    Server: Microsoft-IIS/8.5
    X-AspNet-Version: 4.0.30319
    X-Powered-By: ASP.NET
    Date: Wed, 20 Sep 2017 16:11:06 GMT
    Connection: close
    Content-Length: 522

So the crash is occurring in the line

    headers.put(header.getName(), header.getValue());

Since the `header.getName()` is null, then the `headers.put` throws the `NullPointerException`. I changed the line to

    headers.put(header.getName() != null ? header.getName() : "", header.getValue());

And that crash didn't occur anymore. There were 2 other places I made the same update before my app works properly again. `DiskBasedCache.java` within the function `writeHeaderList` and `BasicNetwork.java` within the function `combineHeaders`.

I'm hoping this is the correct update for this issue.

jpd236:
- NetworkResponse constructors are documented as accepting null maps,
so they should actually do so.
- The new NetworkResponse constructor was ambiguous with the old one
when "null" is provided as the headers, which breaks compile-time
backwards compatibility. Reorder the arguments in the new constructor
to work around this.
- HurlStack previously stripped out headers with null names, which was
critical as HttpUrlConnection returns the status line in this manner.
Reintroduce this filter.

Unit tests added to catch regressions here in the future.

Fixes #95

### DIFF
--- a/src/main/java/com/android/volley/NetworkResponse.java
+++ b/src/main/java/com/android/volley/NetworkResponse.java
@@ -35,7 +35,7 @@ public class NetworkResponse {
      * @param headers Headers returned with this response, or null for none
      * @param notModified True if the server returned a 304 and the data was already in cache
      * @param networkTimeMs Round-trip network time to receive network response
-     * @deprecated see {@link #NetworkResponse(int, byte[], List, boolean, long)}. This constructor
+     * @deprecated see {@link #NetworkResponse(int, byte[], boolean, long, List)}. This constructor
      *             cannot handle server responses containing multiple headers with the same name.
      *             This constructor may be removed in a future release of Volley.
      */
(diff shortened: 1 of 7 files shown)


goghAta:
Please merge it.

PR #98, opened 2017-09-26
Fix an issue when the first request returns an error.

uhager:
The next waiting request was sent to the network without the Listener set. Added Listener for callback to release waiting requests.


jpd236:
Thanks. Is it possible to add a unit test to cover the bug (i.e. one that would fail without the fix?)

uhager:
Added test.

PR #103, opened 2017-09-27
Improve request cancelation handling.

### LINKED ISSUE #15: unableto clear mErrorListener causing memory leaked.
I need to clear error listener after request is cancel (when call onStop()).

Currently that is no way to clear mErrorListener in Request class, which will cause leaked if the error listener implementation is reference fragment or activity.

### LINKED ISSUE #85: RequestFuture#get blocks forever if an in-flight request is cancelled
RequestFuture#get waits for either onResponse or onErrorResponse to be called before attempting to return a result. However, if a request is cancelled, by design, neither callback will be executed. This means get() will block forever (or until the user-specified timeout, if one is given).

Proposal to fix:

- Add an optional CancelListener that can be set on a request
- Have cancel() only set mCanceled = true when called from the main thread, and queue a runnable to do so otherwise
- Call the cancel listener when setting mCanceled = true (on the main thread)
- Have RequestFuture implement CancelListener and set itself as the listener in setRequest, and call notify() in the listener.
- Ensure we use setRequest everywhere we use RequestFuture so it is cancellable

jpd236:
- Provide stronger thread safety guarantees by locking access to any
variable which can be mutated during the processing of a request (i.e.
by the CacheDispatcher/NetworkDispatcher threads) and dispatching
response callbacks while holding the same lock that is held when
setting mCanceled. Unfortunately there's not much we can do about
alternative ResponseDelivery mechanisms, which are hopefully rare.

- Add a cancel listener interface and use it to prevent RequestFuture
from blocking for the complete timeout (or indefinitely) if a request
is canceled.

- Since listeners are often Android component classes like Activitys,
clear them upon cancellation. Unfortunately, this must be done in each
Request subclass to work, assuming the subclass follows the standard
pattern of holding the response listener as a member. But this enables
concerned apps to resolve this leak.

Fixes #15
Fixes #85

### DIFF
--- a/src/main/java/com/android/volley/ExecutorDelivery.java
+++ b/src/main/java/com/android/volley/ExecutorDelivery.java
@@ -88,31 +88,7 @@ public class ExecutorDelivery implements ResponseDelivery {
         @SuppressWarnings("unchecked")
         @Override
         public void run() {
-            // If this request has canceled, finish it and don't deliver.
-            if (mRequest.isCanceled()) {
-                mRequest.finish("canceled-at-delivery");
-                return;
-            }
-
-            // Deliver a normal response or error, depending.
-            if (mResponse.isSuccess()) {
-                mRequest.deliverResponse(mResponse.result);
-            } else {
-                mRequest.deliverError(mResponse.error);
-            }
-
-            // If this is an intermediate response, add a marker, otherwise we're done
-            // and the request can be finished.
-            if (mResponse.intermediate) {
-                mRequest.addMarker("intermediate-response");
-            } else {
-                mRequest.finish("done");
-            }
-
-            // If we have been provided a post-delivery runnable, run it.
-            if (mRunnable != null) {
-                mRunnable.run();
-            }
+            mRequest.deliverResponse(mResponse, mRunnable);
        }
     }
 }
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -56,6 +56,12 @@ public abstract class Request<T> implements Comparable<Request<T>> {
         int PATCH = 7;
     }
 
+    /** Callback interface to notify listeners when a request has been canceled. */
+    public interface CancelListener {
+        /** Called when a request was canceled before the response could be delivered. */
+        void onRequestCanceled();
+    }
+
     /**
      * Callback to notify when the network request returns.
      */
@@ -83,8 +89,11 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     /** Default tag for {@link TrafficStats}. */
     private final int mDefaultTrafficStatsTag;
 
+    /** Lock to guard state which can be mutated after a request is added to the queue. */
+    private final Object mLock = new Object();
+
     /** Listener interface for errors. */
-    private final Response.ErrorListener mErrorListener;
+    private Response.ErrorListener mErrorListener;
 
     /** Sequence number of this request, used to enforce FIFO ordering. */
     private Integer mSequence;
@@ -96,9 +105,11 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     private boolean mShouldCache = true;
 
     /** Whether or not this request has been canceled. */
+    // GuardedBy(mLock)
     private boolean mCanceled = false;
 
     /** Whether or not a response has been delivered for this request yet. */
+    // GuardedBy(mLock)
     private boolean mResponseDelivered = false;
 
     /** Whether the request should be retried in the event of an HTTP 5xx (server) error. */
@@ -118,10 +129,11 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     private Object mTag;
 
     /** Listener that will be notified when a response has been delivered. */
+    // GuardedBy(mLock)
     private NetworkRequestCompleteListener mRequestCompleteListener;
 
-    /** Object to guard access to mRequestCompleteListener. */
-    private final Object mLock = new Object();
+    /** Optional listener for request cancel events. */
+    private CancelListener mCancelListener;
 
     /**
      * Creates a new request with the given URL and error listener.  Note that
@@ -182,6 +194,18 @@ public abstract class Request<T> implements Comparable<Request<T>> {
      */
     public Response.ErrorListener getErrorListener() {
         return mErrorListener;
+    }
+
+    /**
+     * @return this request's {@link CancelListener}, if any.
+     */
+    public CancelListener getCancelListener() {
+        return mCancelListener;
+    }
+
+    /** Sets a {@link CancelListener} which will be notified when the request is canceled. */
+    public void setCancelListener(CancelListener listener) {
+        mCancelListener = listener;
     }
 
     /**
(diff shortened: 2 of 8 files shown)


jpd236:
@googlebot is apparently having issues right now; the above comment can be ignored and I'll request a re-check once it's resolved.

src/main/java/com/android/volley/Request.java
```diff
@@ -56,6 +56,12 @@
         int PATCH = 7;
     }
 
+    /** Callback interface to notify listeners when a request has been canceled. */
+    public interface CancelListener {
```
joebowbeer: The names seems odd. Ideas for a better name:

* OnRequestCanceledListener
* CancellationListener
* OnCancelListener

jpd236:
I appreciate the concern, but given the volume of reports of the leak in #15 and duped reports, I'd like to get this resolved as well in the next release. I also find the bug with RequestFuture hanging forever in wait() to be pretty severe in nature - it caused thread pool exhaustion in one of our internal apps. The DiskBasedCache bug is severe as well when it does happen, but has been present for quite some time already.

This does happen to be the last planned change on the roadmap for 1.0.1, though. I wouldn't anticipate other changes being merged before release unless they fix regressions introduced since 1.0.0. So hopefully we'll have an RC build ready soon.

src/main/java/com/android/volley/Request.java
```diff
@@ -320,17 +344,81 @@ public String getCacheKey() {
...
+     */
+    protected void onCanceled() {
+    }
+
+    /** Deliver the given response as long as the request isn't canceled. */
+    void deliverResponse(Response<T> response, Runnable runnable) {
+        // To uphold the contract of cancel(), we hold the same lock here around both the check of
+        // isCanceled() as well as the actual response delivery. This ensures that after a call
+        // to cancel() returns, we will never deliver the response.
+        synchronized (mLock) {
```
joebowbeer: The change from open calls (without lock held) to closed calls makes me nervous about deadlock.

src/main/java/com/android/volley/ExecutorDelivery.java
```diff
@@ -88,31 +88,7 @@ public ResponseDeliveryRunnable(Request request, Response response, Runnable run
...
-                mRequest.addMarker("intermediate-response");
-            } else {
-                mRequest.finish("done");
-            }
-
-            // If we have been provided a post-delivery runnable, run it.
-            if (mRunnable != null) {
-                mRunnable.run();
-            }
+            mRequest.deliverResponse(mResponse, mRunnable);
```
uhager: Could you call that something different, maybe handleResponse or prepareResponseDelivery of maybeDeliverResponse or something? It was a bit confusing to read for me since there are now 2 deliverResponse with different arguments and having different jobs.

src/main/java/com/android/volley/Request.java
jpd236: CancellationListener makes sense - the intent is to be consistent with "ErrorListener", so the prefixes should both be nouns describing what happened.

src/main/java/com/android/volley/Request.java
uhager: That's a lot of stuff behind this lock, I'm not sure that's needed. I think instead we could add a check whether finish() has been called (addMarker() after finish() leads to IllegalStateException). Since this method (and finish) are package-private it is anyway not possible for users to access them without using ExecutorDelivery. And as long as ExecutorDelivery always runs them on the same thread, we can't have conflicts here as far as I can tell. What do you think?

src/main/java/com/android/volley/ExecutorDelivery.java
jpd236: Yeah, agreed this is confusing; I couldn't come up with a better name. If we end up keeping this (per my other comment) then maybe deliverResponseAndFinish() or something to indicate the additional work that this method is responsible for doing.

src/main/java/com/android/volley/Request.java
```diff
@@ -96,9 +105,11 @@
     private boolean mShouldCache = true;
 
     /** Whether or not this request has been canceled. */
+    // GuardedBy(mLock)
```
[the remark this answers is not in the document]
jpd236: We don't have a dependency on the support lib yet, so for now I'm holding off on this. At some point later on we might add it to cover things like this, @VisibleForTesting, etc.

src/main/java/com/android/volley/Request.java
```diff
@@ -96,9 +105,11 @@
...
     private boolean mCanceled = false;
```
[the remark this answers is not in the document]
jpd236: It's not sufficient as it stands due to holding mLock around the response dispatching below; it might be sufficient if we change that, but I'm not sure that n volatile variables is necessarily more performant than a simple lock, and it's certainly more tricky to reason about :)

src/main/java/com/android/volley/Request.java
```diff
@@ -320,17 +344,81 @@ public String getCacheKey() {
...
+     * guarantees otherwise.
+     *
+     * <p>Subclasses should generally override {@link #onCanceled()} instead to perform an action
+     * when the request is canceled (such as clearing response listeners to avoid leaks).
      */
     public void cancel() {
-        mCanceled = true;
+        synchronized (mLock) {
+            mCanceled = true;
+            onCanceledInternal();
```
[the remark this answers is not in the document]
jpd236: That's a fair point; I could change this to avoid double-cancellation.

src/main/java/com/android/volley/Request.java
jpd236: @joebowbeer I acknowledge the concern, but I don't see a way to properly implement the guarantees that are advertised by cancel() without ensuring that the same lock is held for both cancel() and for the response dispatching itself. Without this lock, it's possible for someone to call cancel() and still have the callbacks called (if we've already checked isCanceled() here).

@uhager It doesn't seem to me like finish() is always called (in the response.intermediate case) so I'm not sure that works in every case, right? And the assumption about ExecutorDelivery running on the same thread is the one this locking is intended to resolve.

Another option might be to weaken the guarantee by stating that you must call cancel() from the same thread as the ResponseDelivery being used (typically the main thread) in order to guarantee that the callbacks aren't invoked. If you call cancel() from another thread, and the main thread is in the process of delivering responses, it may still end up delivering the response. In this case we could leave the dispatching as it was and just accept the risk of a post-cancel dispatch. However, this would mean needing to lock writes to the error listener / response listener.

@jjoslin - any thoughts?

uhager:
@jpd236 I meant that ExecutorDelivery always sends deliverResponse on the same Handler. So deliverResponse can't be called simultaneously, since - as far as I can tell - no other class calls it and it's not accessible to users. Or does RequestFuture change that behaviour in ExecutorDelivery? 
For an intermediate response, we want it to be allowed to be called again, just not after finish was called. So if 2 threads call ExecutorDelivery#postResponse with non-intermediate responses, only the first deliverResponse will run, the second one will see that finish has been called and return. If we allow users to call deliverResponse(response, runnable), or it is called somewhere else in Volley, then I agree, we need some locking.

uhager:
Thinking about this some more, wouldn't it make more sense to synchronize access to mErrorListener? Where it is set to null when cancelled and then in deliverResponse around deliverError? Check if it's non-null and then synchronize using it? 
I'm just worried that this is a very big block that is locked and can block a thread that tries to cancel the request for some time. At the very least we should use a dedicated lock for this rather than reusing the other one.

src/main/java/com/android/volley/Request.java
```diff
@@ -320,17 +344,81 @@ public String getCacheKey() {
...
-        return mCanceled;
+        synchronized (mLock) {
+            return mCanceled;
+        }
+    }
+
+    private void onCanceledInternal() {
+        mErrorListener = null;
+        if (mCancelListener != null) {
+            mCancelListener.onRequestCanceled();
```
jjoslin: The other callbacks run on the same thread as the ResponseDelivery being used whereas this runs on the calling thread. I realize in most cases those threads will be the same (main thread) but what do you think about providing a stronger guarantee about the thread onRequestCanceled() runs on?

src/main/java/com/android/volley/Request.java
jjoslin: AFAICT the same lock needs to be held during cancel() and response dispatching to uphold the cancel() contract as @jpd236 states. I feel like this is a reasonable thing to do but also acknowledge that it does create the potential for deadlock (although I think it will be rare in practice). I support/prefer the current implementation but I wouldn't be opposed to weakening the guarantee either.

src/main/java/com/android/volley/Request.java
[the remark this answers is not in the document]
joebowbeer: Looking at this a little closer, I see that synchronization is being added to methods that are likely to be called on the main thread. This should be avoided. In particular, Request.cancel() is waiting for a lock held by the ExecutorDelivery.

src/main/java/com/android/volley/toolbox/RequestFuture.java
```diff
@@ -51,9 +52,10 @@
  *
  * @param <T> The type of parsed response this future expects.
  */
-public class RequestFuture<T> implements Future<T>, Response.Listener<T>,
-       Response.ErrorListener {
+public class RequestFuture<T> implements Future<T>, Response.Listener<T>, Response.ErrorListener,
```
joebowbeer: I'm not sure how much difference this makes, but the pattern I like to use for classes that implement Future is to have them wrap an internal FutureTask instance to handle the synchronization.  For example, see BackgroundTask in JCiP (Ch. 9 ). This makes it easier to get the synchronization correct, and makes it more apparent that it is.

jpd236:
OK - this PR seems too contentious to try and merge all at once :) Let me try to split it into more reasonable chunks so it's more manageable; I'll send them out separately.

First of all, thanks all for the detailed comments; I really do appreciate it.

For the leak, I need to think it through a bit more but I think you're probably right that we can just synchronize access to mErrorListener/mListener in Request and each of the toolbox subclasses, and then clear them on cancel. This doesn't guarantee that deliverResponse()/deliverError() won't be called after calling cancel(), but calling them should just no-op if the respective listener hasn't been nulled out, so we should be okay just updating the Javadoc to note this limitation - there are no guarantees if cancel() is called from a different thread than the response delivery. It also leaves this locking up to the Request subclass which isn't ideal, but there's not much we can do about that without holding our own lock for code we don't control.

Please note that grabbing a lock on the main thread should not be a problem; it should finish in microseconds. It's only an issue if that lock is contentious or if the work performed inside the lock can take a non-trivial amount of time. Certainly if it's safe to avoid it's better, as with anything on the main thread, but I don't think it's a concern here when necessary.

For the cancel listener - it'll be a bit more involved, but a possible option here is to add a postCancel to ResponseDelivery and plumb cancel events through there, so for our default implementation it'll run on the main thread. The main issue is that ResponseDelivery is an interface, so as we did with the Apache HTTP refactoring for HttpStack, we'd need to add a new interface and continue to support the old one (dropping cancellation events) for anyone using it. I may punt on this for a while given the complexity and uncertainty about the API, and instead just update RequestFuture to warn about using wait(0) in some form.

I think that covers the higher level open questions. For the nitty-gritty we can cover that on each PR.

jpd236:
Leak/thread-safety: https://github.com/google/volley/pull/105

PR #109, opened 2017-10-13
Fixes Unchecked assignment: 'com.android.volley.RequestQueue.RequestFinishedListener' to 'com.android.volley.RequestQueue.RequestFinishedListener<T>

Ericliu001:
previously in RequestQueue.java the methods: add(), finish(), addRequestFinishedListener(), removeRequestFinishedListener() all define their own method level generic types, which makes the type of RequestFinishedListener<> unchecked - reported by compiler.

This commit creates a class level generic type T for RequestQueue.java and complies with PECS (Producer Extends Consumer Super) principles.

### DIFF
--- a/src/main/java/com/android/volley/RequestQueue.java
+++ b/src/main/java/com/android/volley/RequestQueue.java
@@ -38,7 +38,7 @@ public class RequestQueue<T> {
     /** Callback interface for completed requests. */
     public interface RequestFinishedListener<S> {
         /** Called when a request has finished processing. */
-        void onRequestFinished(Request<S> request);
+        void onRequestFinished(Request<? extends S> request);
     }
 
     /** Used for generating monotonically-increasing sequence numbers for requests. */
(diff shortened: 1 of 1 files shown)


jpd236:
I'm not sure I understand what the generic type T is supposed to represent. A RequestQueue inherently supports Requests of various types. It essentially looks like T is effectively "the base class of all requests ever sent to the queue", which doesn't seem like a useful notion to be tracking, because in most cases it would just be Object.

In any event, this doesn't feel like an API compatible change (at least if you have compiler warnings set to break your build).

Ericliu001:
@jpd236 , yeah you are right about the generic type T of RequestQueue. I was just trying to fix the compiler warning in com/android/volley/RequestQueue.java:240, which is due to the method level generic type T not being the same for each method, for example:
The <T> in mehod:
<T> void finish(Request<T> request) 
is different from the <T> in:
public  <T> void addRequestFinishedListener(RequestFinishedListener<T> listener) 
I believe using wild card might be a better choice.
I will update my pull request and please let me know your thoughts, thank you!

jpd236:
At a quick glance it was probably a mistake to have RequestFinishedListener include a generic type for the same reason it would be wrong to do so on RequestQueue. The compiler can't know the type of Request that will be passed to the listener at compile time, so it's only offering an illusion of type safety and there'd presumably be a crash if you had a Request<T> in the queue where T didn't match the RequestFinishedListener's type.

I'd love to fix it, but we can't unless/until we decide to make breaking API changes. So until then suppressing the warning on that cast may be the best we can do :(

Ericliu001:
Sure, thank you for participating in the discussion. Pull Request closed.

PR #112, opened 2017-10-16
Log fix

NavidHt:
Fixed 2 (minor) bugs:
1. In VolleyLog class, where finding the "caller" of log method from the call stack trace.
2. In Request class, where calling MarkerLog.finish() when Request finishes off of the main thread.

### DIFF
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -249,7 +249,7 @@ public abstract class Request<T> implements Comparable<Request<T>> {
                     @Override
                     public void run() {
                         mEventLog.add(tag, threadId);
-                        mEventLog.finish(this.toString());
+                        mEventLog.finish(Request.this.toString());
                     }
                 });
                 return;
--- a/src/main/java/com/android/volley/VolleyLog.java
+++ b/src/main/java/com/android/volley/VolleyLog.java
@@ -35,6 +35,12 @@ public class VolleyLog {
     public static boolean DEBUG = Log.isLoggable(TAG, Log.VERBOSE);
 
     /**
+     * {@link Class#getName()} uses reflection and calling it on a potentially hot code path may
+     * have some cost. To minimize this cost we fetch class name once here and use it later.
+     */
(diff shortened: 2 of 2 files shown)


src/main/java/com/android/volley/Request.java
```diff
@@ -249,7 +249,7 @@ void finish(final String tag) {
                     @Override
                     public void run() {
                         mEventLog.add(tag, threadId);
-                        mEventLog.finish(this.toString());
+                        mEventLog.finish(Request.this.toString());
```
jpd236: Subtle... nice catch!

src/main/java/com/android/volley/VolleyLog.java
```diff
@@ -88,8 +88,8 @@ private static String buildMessage(String format, Object... args) {
         // Walk up the stack looking for the first caller outside of VolleyLog.
         // It will be at least two frames up, so start there.
         for (int i = 2; i < trace.length; i++) {
-            Class<?> clazz = trace[i].getClass();
-            if (!clazz.equals(VolleyLog.class)) {
+            String clazz = trace[i].getClassName();
+            if (!clazz.equals(VolleyLog.class.getName())) {
```
jpd236: Using the class name instead of the class should be reasonable, but the one caveat is that Class#getName() uses reflection, which we should aim to minimize on a potentially hot code path - and this is a loop inside of a logging method.

This cost can be minimized if we fetch the class name in a static constant, e.g.

`private static final String CLASS_NAME = VolleyLog.class.getName()`

and reference it here so it's loaded once per process instead of on every iteration of this loop.

NavidHt:
Changed the code as suggested in code [review](https://github.com/google/volley/pull/112#discussion_r144932826). I never thought that Class.getName() uses reflection.

NavidHt:
Glad to hear that 😊. This would be my first contribution to a well known open source project, though it's just some minor changes.

jpd236:
All merged. Thanks for your patience and contribution.

PR #117, opened 2017-11-08

### LINKED ISSUE #114: Cache/NetworkDispatcher leaking finished requests
Hi Team,

jpd236:
their contents to leak. Clearing the variable explicitly is not a
sufficient fix because the clear is optimized out as it's not
observable. Extracting the loop contents to a helper function is
likely to work but risks being inlined by a code optimizer. So we
poll() once (which is non-blocking) before waiting to ensure that the
previous request is always cleared expediently.

Should contain no functional changes otherwise.

Fixes #114

### DIFF
--- a/src/main/java/com/android/volley/NetworkDispatcher.java
+++ b/src/main/java/com/android/volley/NetworkDispatcher.java
@@ -84,16 +84,21 @@ public class NetworkDispatcher extends Thread {
         Process.setThreadPriority(Process.THREAD_PRIORITY_BACKGROUND);
         while (true) {
             long startTimeMs = SystemClock.elapsedRealtime();
-            Request<?> request;
-            try {
-                // Take a request from the queue.
-                request = mQueue.take();
-            } catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/android/volley/NetworkDispatcher.java
```diff
@@ -84,16 +84,21 @@ public void run() {
...
-                request = mQueue.take();
-            } catch (InterruptedException e) {
-                // We may have been interrupted because it was time to quit.
-                if (mQuit) {
-                    return;
+            // Take a request from the queue. Polling before the call to take is a workaround for an
+            // Android platform issue where the previous request is not GC'd until take() returns,
+            // which can take indefinitely long if no new requests are added to the queue. See also
+            // https://github.com/google/volley/issues/114
+            Request<?> request = mQueue.poll();
```
jpd236: hmm, I guess this runs the risk of regressing if the queue has items when quit() is called, since we'll never notice the interrupt and keep looping. So I'm going to go back to what I did before, even if it's a larger looking CL, to factor this into a helper method.

PR #118, opened 2017-11-08

### LINKED ISSUE #114: Cache/NetworkDispatcher leaking finished requests

jpd236:
their contents to leak. Clearing the variable explicitly is not a
sufficient fix because the clear is optimized out as it's not
observable. The best we can do is extract the contents of the loops in
question to helper methods and hope that they are complex enough that
code optimizers like Proguard won't inline them.

Should contain no functional changes otherwise.

Fixes #114

### DIFF
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -93,82 +93,88 @@ public class CacheDispatcher extends Thread {
 
         while (true) {
             try {
-                // Get a request from the cache triage queue, blocking until
-                // at least one is available.
-                final Request<?> request = mCacheQueue.take();
-                request.addMarker("cache-queue-take");
-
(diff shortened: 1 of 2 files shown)


joebowbeer:
I suspect removing the final is the key change, as the "clearing local variable" issue has not been a problem in years.

The final is needed to preserve the value for the inner class. Converting the inner class to static nested removes need to freeze the value in a final variable, and I think that is enough to nudge the compiler into the right course of action.

The current compilation, with the final, may be a bug (in ART) that should be reported.

jpd236:
It's not the key change, unfortunately, and the small repro case from the bug proves that (also the fact that NetworkDispatcher wasn't final but also has the same problem).

This is quite a can of worms, and we're still looking at the right fix. The bug has a little more context and another example of this happening in the framework. It's been reported but it's not a trivial fix and wouldn't affect backwards compatibility anyway.

Issue #127, opened 2017-12-20
Out of memory with Volley 1.1.0


megataps:
I got this crash when upgrade to version 1.1.0

Fatal Exception: java.lang.OutOfMemoryError: Failed to allocate a 335787708 byte allocation with 1588101 free bytes and 507MB until OOM
       at java.util.ArrayList.(ArrayList.java)
       at com.android.volley.toolbox.DiskBasedCache.readHeaderList(SourceFile:606)
       at com.android.volley.toolbox.DiskBasedCache$CacheHeader.readHeader(SourceFile:437)
       at com.android.volley.toolbox.DiskBasedCache.initialize(SourceFile:170)
       at com.android.volley.CacheDispatcher.run(SourceFile:92)

joebowbeer:
I see now that I overlooked that case. The unprotected ArrayList allocation is causing OOME due to a bogus size in the cache:

`new ArrayList<Header>(size);`

The fix is to not trust the size:

`new ArrayList<Header>();`

Then the subsequent read calls will run off the end of the input stream, which is an exception that will be caught and handled by deleting the corrupted cache entry.

I'll post a fix (and a new unit test!)

joebowbeer:
See #12

In the next version of the on-disk cache, the header size should be persisted as unsigned short..

jpd236:
This is now in 1.1.1-SNAPSHOT builds. It would be excellent if someone encountering this issue could update to 1.1.1-SNAPSHOT and verify the fix. Thanks!

j-t-h:
Do you have an update when 1.1.1 with be released and not in the snapshot builds? I am holding an app release to wait on this fix.

jpd236:
We don't have a planned date at this time - there are [more open issues](https://github.com/google/volley/milestone/4) to investigate first, unless there becomes an urgent need to cut an official release, e.g. for a security issue or major bug. That being said, official releases at this point really are just SNAPSHOT builds which are promoted to official release numbers, so there's not much difference in practice between depending on a particular snapshot and having us designate that particular snapshot as a particular release.

joebowbeer:
I suggest releasing it sooner rather than later. The mcxiaoke folks are also looking for a fix:

https://github.com/mcxiaoke/android-volley/issues/37

If releases are really just promoted snapshots, is there any reason not to release more often? The problem with SNAPSHOT is that it might break the next time something is merged to master.

jpd236:
[...]

PR #131, opened 2017-12-21
Guard against illegal argument and OOM exceptions in readHeaderList

joebowbeer:
Fix for #127

I verified that the two unit tests fail without this PR.


jpd236:
Thanks, change is reasonable and test coverage is solid :)

PR #135, opened 2017-12-25
Rounding corners radius for NetworkImageView

smaslenko:


uhager:
I'm not sure this belongs in the general Volley implementation. Would it be possible to instead add a listener to NetworkImageView that is called in onResponse and can then do whatever image manipulation the user wants?

jpd236:
+1 - I agree this is a bit outside the scope of what Volley should be doing.

As another potential suggestion, you could extend NetworkImageView and override setImageBitmap/setImageResource, and apply your transformation there.

In general, I'd be open to a PR that adds hooks in places that don't have them if it's difficult for an app to implement this kind of behavior, but in this case I think you should have all you need. Let us know if not!

PR #137, opened 2018-01-03
Enable TLSv1.1 and TLSv1.2 by default on API 16-19.

### LINKED ISSUE #77: OS < Android 4.4 doesn't work with TLS 1.2
Details from Mihai Petresco : @mihaicristianpetrescu
I've managed to build an app that fixes the issue for both devices that have/don't have Google Play Services installed.
The fix in our case has forced us to enable TLS v1.1 on our backend server because by using only TLS v1.2 no matter what fix I've tried it wasn't working.
With TLS v1.1 enabled, on devices that run Android 4.4, API calls get through with the fix in place. Without it they don't. You can try it out yourself by using an emulator with Android 4.4 and no Google Play Services installed. If we then disable TLS v1.1 and leave just TLS v1.2 it doesn't work anymore:

javax.net.ssl.SSLHandshakeException: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0xb8765850: Failure in SSL library, usually a protocol error

If there is any way of getting it to work with having only TLS v1.2 enabled on the backend API server and a workaround for older Android running devices, that would be great and ideal, because SSLv3, TLS v1 and TLS v1.1 are not recommended to be used on a secure server. I understood that with the proper implementation TLS v1.1 could be used but still, it would be great to have only TLS v1.2 for all our users.
The sample app is attached. 
[SSLTLSTestApp.zip](https://github.com/google/volley/files/1252695/SSLTLSTestApp.zip)

jpd236:
Per:

https://developer.android.com/reference/javax/net/ssl/SSLSocket.html

These protocols are supported on API 16+, but not enabled by default
until API 20+. This has led to frequent bug reports on such devices
for servers which only support TLSv1.2. As such, if no custom
SSLSocketFactory is provided to HurlStack, we can use our own which
adds these new protocols to whatever is already supported by default.

This should be low risk - we're only enabling new protocols without
disabling old ones, and if there were issues with a server's TLSv1.2
implementation then they'd likely be visible to other clients. As
such, the benefits of enabling these protocols by default should
outweigh the risks of regression for existing Volley users, who would
likely want to resolve the server-side issue anyway, but could still
work around the issue by providing their own SSLSocketFactory
implementation.

Fixes #77

### DIFF
--- a/src/main/java/com/android/volley/toolbox/HurlStack.java
+++ b/src/main/java/com/android/volley/toolbox/HurlStack.java
@@ -68,11 +68,19 @@ public class HurlStack extends BaseHttpStack {
 
     /**
      * @param urlRewriter Rewriter to use for request URLs
-     * @param sslSocketFactory SSL factory to use for HTTPS connections
+     * @param sslSocketFactory SSL factory to use for HTTPS connections. If none is provided, the
+     *                         system default will generally be used. However, support for TLSv1.1
+     *                         and 1.2 will be forcefully enabled on devices which support them even
+     *                         if they are not on by default. If this behavior is undesired, provide
+     *                         a custom SSLSocketFactory.
      */
     public HurlStack(UrlRewriter urlRewriter, SSLSocketFactory sslSocketFactory) {
         mUrlRewriter = urlRewriter;
-        mSslSocketFactory = sslSocketFactory;
+        if (sslSocketFactory != null) {
+            mSslSocketFactory = sslSocketFactory;
+        } else {
+            mSslSocketFactory = TlsEnabledSSLSocketFactory.newTlsEnabledSSLSocketFactoryIfNeeded();
+        }
     }
 
     @Override
--- a/src/main/java/com/android/volley/toolbox/InterceptingSSLSocketFactory.java
+++ b/src/main/java/com/android/volley/toolbox/InterceptingSSLSocketFactory.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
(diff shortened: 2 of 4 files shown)


jpd236:
Hrm, now that I've written this, on closer inspection it seems like OkHttp doesn't do this and I'm feeling more and more hesitant.

jhogervorst:
Thanks for your pull request! While it’s not merged into Volley, I was able to use your changes to support TLS 1.1/1.2 on older Android versions (16–20) in my apps 😄

For others looking into this fix, please note that there seems to be a mistake in [`TlsEnabledSSLSocketFactory.java `](https://github.com/jpd236/volley/blob/250aa2c7496cc83641494d55c711c5fcce50cb04/src/main/java/com/android/volley/toolbox/TlsEnabledSSLSocketFactory.java#L48-49):

```
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN
                && Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT_WATCH) {
            // No special factory needed - TLSv1.1 and 1.2 are unsupported prior to Jelly Bean and
            // enabled by default on KitKat Watch+ devices per:
            // https://developer.android.com/reference/javax/net/ssl/SSLSocket.html
            return null;
        }
```

The if statement seems to do the exact opposite of what’s appropriate, since the fix should be applied for SDKs 16–19. In my apps, I changed the lines as follows:

```
        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN
                || Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT_WATCH) {
```

PR #139, opened 2018-01-08
Prevent duplicated/conflicting HTTP Headers for HurlStack

houtianze:
Summary
=======
It seems a subtle bug to me that `HurlStack` will allow duplicated/conflicting HTTP Headers (specifically, the "Content-Type" header) from being added, this PR eliminate this possibility.

Illustration
=========
The following code will generate HTTP request that have conflicting "Content-Type" headers using the current version of Volley, which will result an HTTP 400 response from the server most of the time:
```java
public class PostRequest extends Request<String> {

  @Override
  public String getBodyContentType() {
    return "text/plain";
  }

  @Override
  public Map<String, String> getHeaders() throws AuthFailureError {
    Map<String, String> headers = super.getHeaders();
    headers.put("Content-Type", "application/json");
    return headers;
  }

  // other codes ...
}
```

Assumptions
=========
I assume that Volley intends to support unique HTTP Headers only in requests, because the return type of `getHeaders()` is `Map<String, String>` instead of (the more generic) `Map<String, List<String>>`.

Justifications
==========
- With the assumptions mentioned, the "normal scenarios" will still work as before, but this change prevents users from accidentally creating duplicated/conflicting HTTP Headers that will almost certainly fail.
- If we check the code of (now deprecated) `HttpClientStack`, it does't have this problem (it in the end calls `setHeaders()` instead of `addHeader()` in `performRequest()`), which is the same behaviour as this change applied to `HurlStack`.
- Adding "Content-Type" header is a common task for programmers, so this can potentially reduce some troubleshooting efforts for those who unknowingly tripped this error.

Background
=========
I encountered this issue while using the swagger generated Android client (I believe this issue https://github.com/swagger-api/swagger-codegen/issues/5575 describes exactly what I faced).

### DIFF
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -93,6 +93,7 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     private RequestQueue mRequestQueue;
 
     /** Whether or not responses to this request should be cached. */
+    // TODO(#190): Turn this off by default for anything other than GET requests.
     private boolean mShouldCache = true;
 
     /** Whether or not this request has been canceled. */
@@ -286,7 +287,18 @@ public abstract class Request<T> implements Comparable<Request<T>> {
 
     /** Returns the cache key for this request. By default, this is the URL. */
     public String getCacheKey() {
-        return getUrl();
+        String url = getUrl();
+        // If this is a GET request, just use the URL as the key.
+        // For callers using DEPRECATED_GET_OR_POST, we assume the method is GET, which matches
+        // legacy behavior where all methods had the same cache key. We can't determine which method
+        // will be used because doing so requires calling getPostBody() which is expensive and may
+        // throw AuthFailureError.
+        // TODO(#190): Remove support for non-GET methods.
+        int method = getMethod();
+        if (method == Method.GET || method == Method.DEPRECATED_GET_OR_POST) {
+            return url;
+        }
+        return Integer.toString(method) + '-' + url;
     }
 
     /**
--- a/src/test/java/com/android/volley/RequestTest.java
+++ b/src/test/java/com/android/volley/RequestTest.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
+import com.android.volley.Request.Method;
 import com.android.volley.Request.Priority;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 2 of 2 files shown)


src/main/java/com/android/volley/toolbox/HurlStack.java
```diff
@@ -92,7 +92,7 @@ public HttpResponse executeRequest(Request<?> request, Map<String, String> addit
         URL parsedUrl = new URL(url);
         HttpURLConnection connection = openConnection(parsedUrl, request);
         for (String headerName : map.keySet()) {
-            connection.addRequestProperty(headerName, map.get(headerName));
+            connection.setRequestProperty(headerName, map.get(headerName));
```
jpd236: FYI, technically this one seems like a no-op, because it's a brand new connection and so the connection will presumably not have any properties set. But it's fine to change. Only the one below where the content-type property is set after this is a problem. But it's fine to change for consistency.

houtianze:
Thanks for the detailed reply. Agree that failing/warning loud would be a better approach, but more code changes are needed for sufficient validations.

For https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/HttpClientStack.java#L123 , I guess it won't have this issue, as though this line calls `addHeader()`, it's right after a new `HttpPost` being created, so this `addHeader()` shouldn't give problem there, and subsequent header handling call `addHeaders()` which indeed call `setHeader()` instead, so it won't result in duplicated headers. Maybe we can also consider change that `addHeader()` call to `setHeader()`, similiarly for consistency.

jpd236:
Good point re: HttpClientStack. In fact what's (IMO) broken is that the header from getHeaders will take precedence over the content-type header, whereas in HurlStack it's the other way around (after your CL at least). There may not be duplicates but a different one will be used depending on which stack you're using which certainly isn't intended.

src/main/java/com/android/volley/toolbox/HurlStack.java
```diff
@@ -254,7 +254,7 @@ private static void addBody(HttpURLConnection connection, Request<?> request, by
         // since this is handled by HttpURLConnection using the size of the prepared
         // output stream.
         connection.setDoOutput(true);
-        connection.addRequestProperty(
+        connection.setRequestProperty(
```
anuj-sahai: Hi @houtianze @jpd236,
There have been a lot of issues / confusion related to devs setting the Content-Type header using the getHeaders method instead of overriding the getBodyContentType (maybe also due to lack of docs at https://developer.android.com/training/volley/request-custom.html). Doing that would result in Volley sending request with two Content-Type headers. Usually, servers don't strictly verify this, but in a current ongoing issue we're having that had been very hard to zero in on, MS Azure servers/gateways decline this with a 400 BAD request.

This change should fix the issue, but the users may be confused as to why the Content-Type header is not being set to what they explicitly passed, if they are overriding getHeaders and not getBodyContentType. Would it make sense to do a precheck here for Content-Type header and inform the users about the same (log messages?) and then go ahead with replacing the header?

anuj-sahai: Ah never mind, just saw the conversation. Moving this comment there.

anuj-sahai:
Hi @jpd236, is there some timeline for this being merged to Volley master or 1.1.1 milestone? Or would it be when the remaining 13 open issues in 1.1.1 get fixed?

jpd236:
Sorry, I botched my attempt to amend to this pull request and can't seem to recover it easily. Reopened as #193.

PR #140, opened 2018-01-09
fix: request time contains RequestQueue blocking time.

dezng:


dezng:
I signed it!

jpd236:
Interesting find... seems to have been this way since 1.0.0 at least. Doesn't seem useful in its current form because take() will block indefinitely until a request actually gets queued.

Note though that despite this, the actual network time is also available via NetworkResponse#networkTimeMs, and I don't think that has an issue like this, though it's only present for requests that actually get a network response.

@uhager I know you're using this field - any concerns about this change from your end? If not we should probably take it (and perhaps update the Javadoc for VolleyError#getNetworkTimeMs which is currently non-existent).

uhager:
I don't think we have ever really looked at the network time for errors, only for responses. 
This change looks good to me.

PR #160, opened 2018-04-05
Add a concept of public APIs to Volley.

### LINKED ISSUE #102: Clearer API enforcement using proguard and/or doclava
Currently Volley release builds aren't proguard-obfuscated/optimized. This means that any public method becomes part of the public API. This can be unfortunate when we want to keep something private to the base package and the .toolbox package without making it a public API, as we have no choice but to copy the method or go ahead and make it public.

There are also a number of ad-hoc unit tests which reflect to access every expected public method and make sure they exist, which is a maintenance burden.

It'd be good to:

- Proguard the builds we can "@hide" public methods we don't want to include in builds
- Auto-generate a file containing the full API "signature" across all of Volley, so we can more clearly monitor it for changes. Fail builds if the signature changes without updating this file. This is similar to the Android API process (the current.txt file).

jpd236:
-All public/protected classes/methods/fields are considered public
 API. VolleyApisTest uses reflection to find all public APIs and
 compare against a golden list so that APIs are not accidentally
 changed in the future.

-A new .internal package can be used for any classes which need to
 be public but should not be depended on by Volley clients. Any
 class in this package will be obfuscated by Proguard and excluded
 from Javadocs in release artifacts.

This is a relatively simple approach compared to the doclava-based
approach used by the support library / Android platform, which isn't
worth the complexity. However, one drawback is that this cannot hide
public/protected parts of classes in the core Volley package.
Unfortunately, Proguard does not appear to support such a directive
(short of manually listing everything that needs to be kept).

Fixes #102

### DIFF
--- a/bintray.gradle
+++ b/bintray.gradle
@@ -23,6 +23,10 @@ task sourcesJar(type: Jar) {
 task javadoc(type: Javadoc) {
     source = android.sourceSets.main.java.srcDirs
     classpath += project.files(android.getBootClasspath().join(File.pathSeparator))
+    // Exclude classes with "Internal" in the name from generated Javadoc, as these are obfuscated
+    // and not part of the Volley API.
+    exclude 'com/android/volley/internal/**'
+
 }
 
 task javadocJar(type: Jar, dependsOn: javadoc) {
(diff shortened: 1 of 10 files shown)


jpd236:
FYI, while I think the unit test for public APIs is definitely reasonable, I'm a bit more on the fence about the .internal package directory. It's harder to enforce correctly, and I'm not sure it really buys us all that much (basically just code sharing of private things between the regular package and toolbox package). If there was a way to hide public methods in the Volley package that would definitely be useful, but as it stands this wouldn't really offer a way to fix bugs like #160 as I'd hoped.

Would be curious about your thoughts there.

PR #161, opened 2018-04-05
Add new JSON Response handler which can send any custom payload in re…

abhi2304:
Add new JSON Response handler which can send any custom payload in request body. My team required this feature in current ongoing project in which REST services was designed to accept url-encoded-form-data as request body but in reply generate JSON Object. 
This code is currently being used in the project, it may be a general requirement. Therefore I posted this code here if found useful you may incorporate this in main branch.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/JsonCustomPayloadRequest.java
+++ b/src/main/java/com/android/volley/toolbox/JsonCustomPayloadRequest.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2011 The Android Open Source Project
+ *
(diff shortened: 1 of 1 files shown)


jpd236:
Thanks. However, this seems a bit special case and therefore not really suitable for the general Volley repository - in general I'd expect most servers which return JSON to also expect it in the request.

FWIW, it would probably save you some boilerplate to extend JsonRequest<JSONObject> instead. The constructor takes any string request body - it doesn't matter if it's not valid JSON, because it just gets serialized to bytes - and you can still override getBodyContentType() to return a custom content type other than application/json.

PR #165, opened 2018-04-11
Apply google-java-format to all code.

### LINKED ISSUE #99: google-java-format all code; enforce format for future commits
(Forking from https://github.com/google/volley/issues/1)

- Do a one-time pass using google-java-format to format the Volley codebase according to a consistent/reasonable style guide.

- Add gradle tasks to reformat the tree and validate that the tree is formatted; run the latter as part of CI builds so PRs can be flagged when they aren't formatted. There is at least [one](https://github.com/sherter/google-java-format-gradle-plugin) existing plugin that claims to do this which we might be able to use.

jpd236:
Also adds Gradle commands to format the code and verify formatting on
all files in the tree. CONTRIBUTING.md has been updated with the
steps to format + verify changes, and the format verification has
been added to the continuous integration setup (but if it fails, the
rest of the build will still be attempted to minimize churn).

Fixes #99

### DIFF
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -21,9 +21,7 @@ import android.net.Uri;
 import android.os.Handler;
 import android.os.Looper;
 import android.text.TextUtils;
-
 import com.android.volley.VolleyLog.MarkerLog;
-
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.Collections;
@@ -36,14 +34,10 @@ import java.util.Map;
  */
 public abstract class Request<T> implements Comparable<Request<T>> {
 
-    /**
-     * Default encoding for POST or PUT parameters. See {@link #getParamsEncoding()}.
-     */
+    /** Default encoding for POST or PUT parameters. See {@link #getParamsEncoding()}. */
     private static final String DEFAULT_PARAMS_ENCODING = "UTF-8";
 
-    /**
-     * Supported request methods.
-     */
+    /** Supported request methods. */
     public interface Method {
         int DEPRECATED_GET_OR_POST = -1;
         int GET = 0;
@@ -56,9 +50,7 @@ public abstract class Request<T> implements Comparable<Request<T>> {
         int PATCH = 7;
     }
 
-    /**
-     * Callback to notify when the network request returns.
-     */
+    /** Callback to notify when the network request returns. */
     /* package */ interface NetworkRequestCompleteListener {
 
         /** Callback when a network response has been received. */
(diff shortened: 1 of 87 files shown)


src/main/java/com/android/volley/Request.java
```diff
@@ -245,13 +227,14 @@ void finish(final String tag) {
...
                 Handler mainThread = new Handler(Looper.getMainLooper());
-                mainThread.post(new Runnable() {
-                    @Override
-                    public void run() {
-                        mEventLog.add(tag, threadId);
-                        mEventLog.finish(Request.this.toString());
-                    }
-                });
+                mainThread.post(
+                        new Runnable() {
```
jjoslin: What's the stance on lambdas?

src/test/java/com/android/volley/toolbox/AdaptedHttpStackTest.java
```diff
@@ -1,11 +1,20 @@
 package com.android.volley.toolbox;
 
-import android.util.Pair;
+import static org.junit.Assert.assertEquals;
```
jjoslin: There's some inconsistency (not from this change) with whether to not to use wildcards. I vote for removing the wildcards in favor of explicit imports.

src/main/java/com/android/volley/Request.java
jpd236: I'm hesitant to use them right now just because of how many places Volley code is exported compiled. It *might* work - we'd need recent Gradle plugins which can desugar the Lambdas, and that desugar library would need to work on minSdkVersion 8+. (Or, we'd need to bump Volley's minSdkVersion). For now... I think we should just avoid them.

src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -136,24 +128,32 @@ public NetworkResponse performRequest(Request<?> request) throws VolleyError {
                 if (statusCode == HttpURLConnection.HTTP_NOT_MODIFIED) {
                     Entry entry = request.getCacheEntry();
                     if (entry == null) {
-                        return new NetworkResponse(HttpURLConnection.HTTP_NOT_MODIFIED, null, true,
-                                SystemClock.elapsedRealtime() - requestStart, responseHeaders);
+                        return new NetworkResponse(
+                                HttpURLConnection.HTTP_NOT_MODIFIED,
+                                null,
```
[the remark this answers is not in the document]
jpd236: Agreed, but I'm not sure if there's an easy way to enforce it, though we could do a one-off cleanup at least.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -505,7 +510,7 @@ public int read(byte[] buffer, int offset, int count) throws IOException {
             return result;
         }
 
-        //VisibleForTesting
+        // VisibleForTesting
```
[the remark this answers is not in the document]
jpd236: I'd love to do this, but that would require a new dependency. That being said, android-support-annotations is probably reasonably well enough isolated from the rest of the support library that it could be reasonable, especially if the dependency didn't propagate to clients (e.g. compile time but not runtime).

src/test/java/com/android/volley/toolbox/AdaptedHttpStackTest.java
jpd236: agreed

jpd236:
Filed https://github.com/google/volley/issues/166 to investigate what we can do for each of your comments. Thanks!

PR #185, opened 2018-05-15
Add @Nullable annotations where necessary.

hackbar:
The docstring and code already behave like these params can be null.


src/main/java/com/android/volley/toolbox/ImageRequest.java
```diff
@@ -76,7 +78,7 @@ public ImageRequest(
             int maxHeight,
             ScaleType scaleType,
             Config decodeConfig,
-            Response.ErrorListener errorListener) {
+            @Nullable Response.ErrorListener errorListener) {
```
jpd236: should the listener arg be @Nullable too?

hackbar: The current API is listener cannot be null in the constructor (according to the docstring), but the member variable mListener can be set to null by cancel(). So the member is annotated, but the constructor arg is not.

Issue #145, opened 2018-02-20
Failure to Cache New Response When Cache is Full


abbasshah17:
I have noticed that while Volley successfully caches response it does not however write over the existing old response.

Say, Volley's cache size is 5 MB and all of it is being used. Now if I make a new call to a unique link, volley does not cache the new response.

In my case I have enabled cache and am looking for 304 (`HTTP_NOT_MODIFIED`) and because the latest response is never cached I always get 200 (`HTTP_OK`) even when file/response has changed from server.

How I initialize the Queue:

    mRequestQueue = Volley.newRequestQueue(mContext);

How I add network request to the queue.

    networkRequest = new CustomNetworkResponseRequest(Request.Method.GET, url, listener, errorListener);
	if (tag != null)
		networkRequest.setTag(tag);
	networkRequest.setRetryPolicy(new DefaultRetryPolicy(2500, 1, DefaultRetryPolicy.DEFAULT_BACKOFF_MULT));
	networkRequest.setShouldCache(true);
	requestQueue.addToRequestQueue(networkRequest);

My `CustomNetworkResponseRequest`

    class CustomNetworkResponseRequestextends Request<NetworkResponse>
	{
		private final Response.Listener<NetworkResponse> mListener;
	
		CustomNetworkResponseRequestextends (int method, String url,
									 Response.Listener<NetworkResponse> listener,
									 Response.ErrorListener errorListener)
		{
		    super(method, url, errorListener);
		    mListener = listener;
		}

		@Override
		protected void deliverResponse(NetworkResponse response) 
		{
		    mListener.onResponse(response);
		}
	
		@Override
		protected Response<NetworkResponse> parseNetworkResponse(NetworkResponse response) 
		{
		    return Response.success(response, HttpHeaderParser.parseCacheHeaders(response));
		}
	}

The only way I have found it to work (just for me) is to increase the cache size. While it promises to ensure **I** always get 304 when I should (I don't have large responses or too many calls). Volley should be able to overwrite the existing cache to include the latest calls while discarding oldest.

jpd236:
In theory this should be happening in DiskBasedCache#pruneIfNeeded which is called at the beginning of every put call.

Would you be able to attach the output of "adb logcat | grep Volley" or something similar after running "adb shell setprop log.tag.Volley VERBOSE" and executing a request that you expect should be cached (expiring old entries) but wasn't?

uhager:
[...]

PR #196, opened 2018-05-31
Enforce disk size limit more strictly in DiskBasedCache.

### LINKED ISSUE #57: DiskBasedCache uses different inputs for file size
Forking from https://github.com/google/volley/pull/52#issuecomment-315856441:

Some parts of DiskBasedCache use the file size as the size of the cache entry, which includes the header overhead. Other parts use the data size which doesn't include the header overhead.

We should make this sane. Some constraints:

* I'd generally prefer we use an accurate size on disk since we advertise the cache as using at most that size, but not including the overhead means we might exceed it.
* It's currently hard to calculate the size prior to writing the entry because we don't know how big the CacheHeader will be. I think it would be technically feasible to calculate this deterministically though given that serialization format is fixed (or at least to estimate it...)

jpd236:
-Previously, when new entries were written to the cache with put(),
the size accounted for in those entries did not include cache overhead
when actually storing them on disk. That overhead is now included.

-We now prune excess entries after writing new entries to disk instead
of before (so that we know the size-on-disk). This prevents the cache
from exceeding the size limit due to one large entry (until the next
put). The cache will now exceed the maximum size when writing an entry
which pushes it over the limit; however, this excess will be cleared
before put() returns rather than some indefinite time in the future.

Fixes #57

### DIFF
--- a/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
@@ -65,7 +65,7 @@ public class DiskBasedCache implements Cache {
     private static final int DEFAULT_DISK_USAGE_BYTES = 5 * 1024 * 1024;
 
     /** High water mark percentage for the cache */
-    private static final float HYSTERESIS_FACTOR = 0.9f;
+    @VisibleForTesting static final float HYSTERESIS_FACTOR = 0.9f;
 
     /** Magic number for current version of cache file format. */
     private static final int CACHE_MAGIC = 0x20150306;
@@ -74,7 +74,9 @@ public class DiskBasedCache implements Cache {
      * Constructs an instance of the DiskBasedCache at the specified directory.
      *
      * @param rootDirectory The root directory of the cache.
-     * @param maxCacheSizeInBytes The maximum size of the cache in bytes.
+     * @param maxCacheSizeInBytes The maximum size of the cache in bytes. Note that the cache may
+     *     briefly exceed this size on disk when writing a new entry that pushes it over the limit
+     *     until the ensuing pruning completes.
      */
     public DiskBasedCache(File rootDirectory, int maxCacheSizeInBytes) {
         mRootDirectory = rootDirectory;
@@ -166,8 +168,6 @@ public class DiskBasedCache implements Cache {
                                 new BufferedInputStream(createInputStream(file)), entrySize);
                 try {
                     CacheHeader entry = CacheHeader.readHeader(cis);
-                    // NOTE: When this entry was put, its size was recorded as data.length, but
-                    // when the entry is initialized below, its size is recorded as file.length()
                     entry.size = entrySize;
                     putEntry(entry.key, entry);
                 } finally {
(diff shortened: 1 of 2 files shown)


jpd236:
Thanks for your comments!

Re: 1 - there is absolutely no guarantee that a new entry won't get removed. It will be removed if it either takes over 100% of the cache size on its own, or over 90% of the cache size if other entries were already present (due to hysteresis). To me it seems like a bug that you could put in an entry that was larger than the requested maximum size (with no size limit at all!) and have it stick around until the next put() call. Why expect such a large entry to be retained?

Re: 2 - Yeah, that's fair. It might have been relevant to whatever was happening in https://github.com/google/volley/issues/145 but we never really got to the bottom of that. Moving this to the 1.2.0 milestone without evidence of a user hitting bugs due to the current behavior.

joebowbeer:
To be clear, the situation that I most want to avoid is a new entry being added, at great cost, and then immediately pruned, along with everything else. Hysteresis is supposed to prevent thrashing.

If you move the prune to the end of the put method, as you have, I think you'll want to switch the hysteresis as well. That is, enter prune if above hysteresis mark, but leave as soon as strict max size is met. This should avoid deleting the entry that was just added: provided it is smaller than strict max size. (Except perhaps for corner cases involving small cache size and large header size?)

jpd236:
I think this may not have been clear - the cache data structure is a LinkedHashMap, which preserves insertion order. So new entries are removed in the order they were inserted, and the newly inserted entry is going to be the last entry we remove, and we'd only remove it if it alone were over the hysteresis mark (90% of the cache size), as we'd remove every other entry first before looking at the latest entry. 

I don't believe there should be significant churn here as a result of this change compared to what we had before, assuming reasonably sized cache entries relative to the maximum size.

joebowbeerxealth:
I understand the LRU nature.

The 'bug' is that an excessive large entry can be cached and it won't be removed until the next entry is written.

However, in this same scenario, your solution will clear the entire cache and leave nothing, which is also not good.

I think there's a middle ground that works better: don't cache if the entry.data.length is larger than some threshold (max size? hysteresis?) Otherwise do cache but don't prune the entry that we just written.

I do think it is better to prune before writing a new entry. This minimizes the peak cache size, and also preserves the newest entry.

In short, a better solution is to simply test entry.data.length at the top of the put method.

jpd236:
Yeah, I was considering an up-front check to make sure the new entry being written isn't too large before writing it. I was leaning against it because entry.data.length is technically the wrong length to evaluate (it's not counting the header overhead), and so to get it precisely right would require an up-front calculation of the header size which is what I was hoping to avoid.

If we just use entry.data.length, we might write an entry only to find that with the cache overhead, it's now too large to fit. But perhaps that's reasonable enough - it would allow us to eliminate caching for entries which are definitely too big, while doing a bit more work for entries which are right at the boundary.

jpd236:
PTAL - I added a check to put which prevents writing a cache entry which would be pruned immediately. This reduces the likelihood of churn, although it's still possible in some edge cases because we don't include the header side in that calculation. I think this is a reasonable compromise, since ultimately this is just an optimization to avoid some churn, and calculating the header size by itself is non-trivial, although feasible. Let me know if you disagree.

PR #206, opened 2018-07-04
FEEDBACK NEEDED Add new request builder

dylan-chong:
A few months ago, I added a new request builder class for several reasons (I will describe these reasons below). I haven't been able to do anything with the changes until recently. Over the past few months, I have gotten quite a few merge conflicts. Therefore I am making this pull request to get some feedback about the core idea of this improvement - do the maintainers actually want this sort of change? And if the maintainers do actually want the sort of change, do they have any ideas on what could be done better (at a high level)?

# Reasons for change
 
So there were a few reasons why I decided to add the request builder.

1. The request body and response must be in the same data format. For example,
   when `POST`ing a `JSONArray`, the response type must be a `JSONArray` --- it
   cannot be a `JSONObject` response. To be able to do this, one would have to
   duplicate the `JsonObjectRequest` class, and modify the constructor. Admittedly, most requests would not encounter this restriction, but it is an annoying restriction nonetheless.

1. It is hard to create a default configuration for all requests. If there are
   `n` `Request` subclasses, then one would have to make `n` factory methods.
   Each factory method would have to create the request configure it in the
   same way as the other factory methods, and then return it. This clearly will
   result in a lot of boilerplate code. Default configuration is very useful for setting default headers

1. An entire new request subclass will need to be created to parse a response in an
   unsupported format --- even if the custom format is extremely simple. This
   requires a new file, or an anonymous class, which requires an extra five
   lines of code. This is another boilerplate code problem. Users probably wouldn't find this a problem most of the time, but it would be nice for the library to have a convenient way to do.

1. The API for configuring `Request`s is split across three places --- the
   constructor, setter methods, and overridable protected methods. This
   inconsistency makes it extremely hard to figure out how to configure the
   `Request`. I've had lots of trouble in the past trying to find out how to configure the request. Very annoying!

1. Because the `Request` subclasses' constructors are used for required
   configuration, there can be a lot of parameters. The `ImageRequest`
   constructor takes seven parameters, and code that uses these constructors
   are hard to read. And since we are using anonymous class is to define the callbacks, constructor calls get very long

# Changes made

- Added a new `RequestBuilder` class (with documentation that I need to update).
- Added `Bodies` and `ResponseParsers` classes were static factory methods, the results of which are to be passed into their respective request builder methods. An important thing to note with this change is that I had to move a lot of code out of the existing request classes into the new ones, as they are now on a more convenient place. I have been careful to not break the existing API, so the methods that I took the code from simply to delegate to the new place and so behave exactly the same as they did before all of these changes. There are also some constants I had to make public in order to do this - this is something I would like to check is okay.
- Added a new `BuildableRequest`, that is packaged private, whose only purpose is to make the request builder possible. 
- Added unit tests for the request builder

# Some notes when looking at the changes

Remember that I just want brief feedback on the core idea of this change - adding a request builder. I haven't completely finished making may changes, there is documentation that needs updating, and some merge conflicts that I have yet to resolve. I still need to do some self code review at some point as well. So please refrain from doing any low level code review.


dylan-chong:
@googlebot I am having some trouble signing the agreement. The email I used to make the commits (which matches my GitHub account) is not a Google account. I also cannot add it as an alternative email to my existing Google account (an alert pop-up says that the emails already linked), presumably because I already have it set up as a recovery email for several accounts. Is it a problem that the MLA used to make the commits does not match the one I'm signing the agreement with

jpd236:
Thanks for sending this out for discussion - I appreciate having a chance to weigh in earlier on!

First, to respond to the initial reasoning for this:

1. This would be pretty easy to resolve in isolation by adding a new constructor to JsonArrayRequest which takes a JSONObject for the request body. Ultimately it's just passed as a string so the type of the data isn't important. I just don't think anyone's asked for this.

2. Most apps I've observed using Volley tend to define a base subclass of Request for common configuration and extend it as needed, which is where you can set common headers or other parameters. But I'll admit that having to use a class hierarchy for this instead of a configurable set of parameters can be constraining.

3. I'm not totally convinced here because "an entire new Request subclass" is really just two methods - parsing and delivering the response, both of which are abstract because they depend on the response format. We could probably reduce this to just the parsing with a base class that takes the Listener in the constructor, which is honestly probably how Request should have worked in the first place, but is hard to change now.

4. Totally agreed. Largely this is legacy cruft due to Volley's long-standing decision not to break existing APIs.

5. Also agreed.

Overall, I think it is an interesting idea to define a new builder interface on top of Request for the usual reasons that builders are considered useful for constructing complex objects - to avoid having constructors with lots of arguments and/or lots of setters. In terms of whether this belongs in Volley proper, I think it has potential, but I'd probably like to see it proven out as a separate library first where it can be iterated on independently of Volley itself. I don't believe you need access to Volley's internals to be able to build an interface like this to simplify Request construction, but let me know if that's not the case in any particular situations.

Adding @jjoslin for any further thoughts he might have.

dylan-chong:
Thanks for the great feedback!

> 1. This would be pretty easy to resolve in isolation by adding a new constructor to JsonArrayRequest which takes a JSONObject for the request body. Ultimately it's just passed as a string so the type of the data isn't important. I just don't think anyone's asked for this.

I suppose adding a new constructor would work. And I'm not surprised no one's asked for this, but it is a symptom of the inflexible inheritance-based design. Oh well this isn't a huge deal anyway

> 2. Most apps I've observed using Volley tend to define a base subclass of Request for common configuration and extend it as needed, which is where you can set common headers or other parameters. But I'll admit that having to use a class hierarchy for this instead of a configurable set of parameters can be constraining.

Yup, but I avoided defining my own request subclass because then it would make the `JSONObjectRequest` class redundant, and other classes. I ended up making subclasses for `JSONObjectRequest` and `JSONArrayRequest`, which were very similar.

> 3. I'm not totally convinced here because "an entire new Request subclass" is really just two methods - parsing and delivering the response, both of which are abstract because they depend on the response format. We could probably reduce this to just the parsing with a base class that takes the Listener in the constructor, which is honestly probably how Request should have worked in the first place, but is hard to change now.

On second thought, I suppose this is not really that important, but the most flexible, ideal API would allow for a single line like

     .parseResponse(ResponseParsers.withString(string -> XML.parse(string))

or something like that, in case you only use this once in your entire code base. Admittedly, this is probably not that useful for most people.

But the annoying thing about creating a new request subclass is that you have to copy in code to configure the default headers and other default stuff (that is if you create request subclasses to put your default configuration in).


>  I don't believe you need access to Volley's internals to be able to build an interface like this to simplify Request construction,

I had to move a lot of the code from within volley is request subclasses into the new `ResponseParsers` and `Bodies` and in doing so had to change the visibility of stuff like constants. So the change can't be done externally, and if you do some ugly copy paste

jpd236:
For what it's worth, I don't feel that duplicating some of what the Json*Request classes are doing is that problematic. I think they're pretty simple classes, and I also suspect that many clients in the wild aren't using it because the JSONObject interface is really unfriendly to use for any non-trivial JSON formats, and so end up building their own Request subclasses off of GSON or similar libraries. So I'd treat them as example requests; if they work for you, great, but if not, it's not really a huge deal to ignore them and build your own layer (and at that point, you can put common logic in a base class).

(Volley doesn't include a GsonRequest or equivalent to avoid pulling in unnecessary dependencies for apps that don't need them, but I've often thought it would make sense to have a separate volley-gson dependency for apps that want it...)

In any event, I may be open to refactoring that enables you to access things that are currently internal if it makes building this as an external library easier, depending on what those are. Looking at ResponseParsers, though, I just don't know that the logic here is all that interesting to be worth avoiding duplication (again, keeping in mind that people parsing JSON often want to use a library like GSON rather than deal with low-level JSONObjects). But happy to provide feedback on any concrete proposals you have for code you'd like to have access to.

dylan-chong:
For what it's worth, I don't feel that duplicating some of what the Json*Request classes are doing is that problematic. I think they're pretty simple classes, and I also suspect that many clients in the wild aren't using it because the JSONObject interface is really unfriendly to use for any non-trivial JSON formats, and so end up building their own Request subclasses off of GSON or similar libraries. So I'd treat them as example requests; if they work for you, great, but if not, it's not really a huge deal to ignore them and build your own layer (and at that point, you can put common logic in a base class).

@jpd236 I see your point. However there is one class that does have some decent logic, `ImageResponseParser` which was taken from `ImageRequest`.

But anyhow, it looks like other people have beaten me to this change. Since they have completed their changes, I think I might as well close this pull request and cut my losses timewise :/

- https://github.com/naver/volley-extensions/tree/master/volleyer
- https://github.com/yakivmospan/volley-request-manager

Issue #209, opened 2018-07-24
ImageListener.onResponse is not been called in some cases


forceLain:
Steps to repro:
1. Initialize a basic Volley request queue`Volley.newRequestQueue(this)`
2. Initialize ImageLoader with no cache `imageLoader = ImageLoader(queue, NoImageCache())`
```
class NoImageCache : ImageLoader.ImageCache {
    override fun getBitmap(url: String?): Bitmap? = null
    override fun putBitmap(url: String?, bitmap: Bitmap?) {}
}
```
3. Create list of items using RecyclerView and NetworkImageView as its items. For each image view, use the same URL.
4. Scroll the list fast

Result:
Some image views are not loading

Sample project: https://github.com/forceLain/VolleyResponseBug

forceLain:
![volley-list-2](https://user-images.githubusercontent.com/3397249/43139580-52f776e4-8f7c-11e8-908e-8b7c33360587.gif)

forceLain:
You can find a sample project here: https://github.com/forceLain/VolleyResponseBug
I feel that creating an array for each result might be not that efficient. Feel free to give any suggestions, I'll be happy to update the fix.

jpd236:
Thanks for the explanation and sample project!

I think there's a simpler and more efficient fix here; there's no reason to even start the second request if the first one is done and just pending batch delivery. We can just add the second request to the first request's batch delivery list. Please take a look at #212 - I verified it resolves the issue in your sample project but would be interested in knowing if I've misunderstood the problem or if there's a problem with that approach.

forceLain:
Sorry for the delay, your changes fix the issue, thanks!

PR #210, opened 2018-07-24
keep all references to BatchedImageRequest

forceLain:
`ImageLoader.mBatchedResponses` can only hold one reference to a `BatchedImageRequest` object. There is a case when `mBatchedResponses` is keeping a `BatchedImageRequest` object to deliver (`mHandler`'s runnable has not been started yet) and another `onResponse`  callback with the same cache key is being called and it is replacing the active `BatchedImageRequest` object with a new one. The original `BatchedImageRequest` is being removed from `mBatchedResponses` so its callback will never be called.

Fixes https://github.com/google/volley/issues/209


jpd236:
I think #212 is a more efficient and simpler fix for this issue. I don't think we shouldn't start a second request for a given cache key if a previous request is complete and just pending batch delivery.

PR #217, opened 2018-08-22
Update NetworkImageView.java

ugokoli:
Added an imageLoader response listener that triggers when image is loaded or not, onSuccess or onError.

My use case:
I am adding blurry effect to the loaded image, but I need to know when the image is downloaded and loaded to view before I can add the blurry effect to the image.

This update makes it possible with example below.

```
//set listener to view
imageView.setImageUrl(resourceUrl, imageLoader, new NetworkImageView.OnResponseListener() {
    @Override
     public void onSuccess(Bitmap bitmap) {
        //Todo: use the image resource
    }

    @Override
     public void onError(VolleyError error) {
        //Todo: handle error gracefully
    }
});
```

### DIFF
--- a/src/main/java/com/android/volley/toolbox/NetworkImageView.java
+++ b/src/main/java/com/android/volley/toolbox/NetworkImageView.java
@@ -40,7 +40,7 @@ public class NetworkImageView extends ImageView {
     /** Current ImageContainer. (either in-flight or finished) */
     private ImageContainer mImageContainer;
     
-    /** ImageLoader response listener interface */
+    /** ImageLoader response listener interface. */
     private ResponseListener mResponseListener;
 
     public NetworkImageView(Context context) {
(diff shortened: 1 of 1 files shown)


jpd236:
If you're trying to apply a transformation to the image upon download, I don't think you should need to make changes to NetworkImageView. Instead, you can override setImageBitmap, which will be called with the downloaded bitmap, and apply the transformation then. Is there any reason that doesn't work for your needs?

ugokoli:
No, that doesn't work for my needs. I need an implementation that is out of the box and part of volley to avoid extra class implementation, inheritance and all those boiler plate codes.

jpd236:
I'm afraid that's not really a strong enough argument for us to make this part of Volley's core library. We try to keep the library to a reasonable minimum - more code in the library is more code that we need to support for all general use cases for all time, which constrains our ability to evolve and grow the library. It's one thing if you cannot accomplish the task you need without changing the core library, but I see no difference between passing a completion callback to setImageUrl vs. making a subclass of NetworkImageView to meet your needs.

PR #227, opened 2018-10-11
Add a default image bitmap and an error image bitmap.

aaijazi:
This lets users choose local bitmaps in addition to local resource IDs.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
@@ -400,7 +400,7 @@ public class DiskBasedCache implements Cache {
                 long softTtl,
                 List<Header> allResponseHeaders) {
             this.key = key;
-            this.etag = ("".equals(etag)) ? null : etag;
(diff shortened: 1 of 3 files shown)


src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -400,7 +400,7 @@ private CacheHeader(
                 long softTtl,
                 List<Header> allResponseHeaders) {
             this.key = key;
-            this.etag = ("".equals(etag)) ? null : etag;
+            this.etag = "".equals(etag) ? null : etag;
```
aaijazi: This ended up being necessary to build and pass tests locally

jpd236: Sorry about that - forked off https://github.com/google/volley/pull/229 to keep the commits separate. Evidently our dependency on ErrorProne is not pinned to a specific version. I tried to pin it but it'd require migrating to a newer version of the plugin which appears to be more involved.

Issue #177, opened 2018-05-04
Fatal Exception: No NameTypeIndex match for SHORT_STANDARD


TuanNA-MTITech:
I could not reproduce, but I usually got crash message from Fabric Crashlytics Report.

Device:
* Nexus 5X and Nexus 6P. OS version 8.1.0

`Fatal Exception: java.lang.AssertionError: No NameTypeIndex match for SHORT_STANDARD
       at android.icu.impl.TimeZoneNamesImpl$ZNames.getNameTypeIndex(TimeZoneNamesImpl.java:724)
       at android.icu.impl.TimeZoneNamesImpl$ZNames.getName(TimeZoneNamesImpl.java:790)
       at android.icu.impl.TimeZoneNamesImpl.getTimeZoneDisplayName(TimeZoneNamesImpl.java:183)
       at android.icu.text.TimeZoneNames.getDisplayName(TimeZoneNames.java:261)
       at java.text.SimpleDateFormat.subFormat(SimpleDateFormat.java:1296)
       at java.text.SimpleDateFormat.format(SimpleDateFormat.java:1004)
       at java.text.SimpleDateFormat.format(SimpleDateFormat.java:974)
       at java.text.DateFormat.format(DateFormat.java:341)
       at com.android.volley.toolbox.HttpHeaderParser.formatEpochAsRfc1123(SourceFile:152)
       at com.android.volley.toolbox.BasicNetwork.getCacheHeaders(SourceFile:256)
       at com.android.volley.toolbox.BasicNetwork.performRequest(SourceFile:130)
       at com.android.volley.NetworkDispatcher.processRequest(SourceFile:120)
       at com.android.volley.NetworkDispatcher.run(SourceFile:87)`

pls help do investigation on that crash.

jpd236:
My theory here would be corruption of the filesystem. If so there's not a ton Volley can do about this. All we're doing is calling SimpleDateFormat with a valid date string and date and that should work, and putting a try-catch around this doesn't seem right for a broken device (as opposed to a broken family of devices, and this is definitely not generally broken on all 5X/6Ps).

Does Crashlytics give you the number of impacted devices?

TuanNA-MTITech:
There are 3 devices. 
I could not reproduce that bug. It's sometime happens or always (not sure) on those devices. All of them are 8.1.0

[...]

PR #231, opened 2018-10-15
Add workaround for AssertionError at BasicNetwork

yjaeseok:
We have some issue about getting date from cache entry.
This is ASOP known issue, but I would like to add some workaround
until that issue is fixed.

Please refer to
https://issuetracker.google.com/issues/110848122

ISSUE=#177

### DIFF
--- a/src/main/java/com/android/volley/toolbox/BasicNetwork.java
+++ b/src/main/java/com/android/volley/toolbox/BasicNetwork.java
@@ -264,8 +264,14 @@ public class BasicNetwork implements Network {
         }
 
         if (entry.lastModified > 0) {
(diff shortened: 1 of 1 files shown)


jpd236:
There are now more details on the bug you linked about the nature of the issue. I think the workarounds described there (in particular calling new Date().toString() before any threads are created) are probably most appropriate here since they prevent the crash in the first place instead of silently working around it and potentially causing Volley to do something unexpected due to ignoring a date header. Unfortunately, I don't think Volley can contain the workaround since there's no guarantee that it is initialized before any threads are created in the app. You would have to add this to Application#onCreate.

PR #238, opened 2018-10-31
Allow tracking of the request's life cycle

artikz:
Add request events listener which will be called on the request life
cycle events (queued, processing started, etc). Also add the ability to
associate data with the requests, without using other data structures,
which is useful when processing request life cycle events.

### DIFF
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -28,6 +28,7 @@ import com.android.volley.VolleyLog.MarkerLog;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -124,6 +125,8 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     @GuardedBy("mLock")
     private NetworkRequestCompleteListener mRequestCompleteListener;
 
+    private Map<String, Object> mTags = new HashMap<>();
+
     /**
      * Creates a new request with the given URL and error listener. Note that the normal response
      * listener is not provided here as delivery of responses is provided by subclasses, who have a
@@ -174,6 +177,19 @@ public abstract class Request<T> implements Comparable<Request<T>> {
      */
     public Object getTag() {
         return mTag;
+    }
+
+    public Request<?> setTag(String key, Object tag) {
+        mTags.put(key, tag);
+        return this;
+    }
+
+    public Object getTag(String key) {
+        return mTags.get(key);
+    }
+
+    public Object removeTag(String key) {
+        return mTags.remove(key);
     }
 
     /** @return this request's {@link com.android.volley.Response.ErrorListener}. */
@@ -248,6 +264,12 @@ public abstract class Request<T> implements Comparable<Request<T>> {
 
             mEventLog.add(tag, threadId);
             mEventLog.finish(this.toString());
+        }
+    }
+
+    void sendEvent(RequestQueue.RequestEvent event) {
+        if (mRequestQueue != null) {
+            mRequestQueue.sendRequestEvent(this, event);
         }
     }
 
--- a/src/main/java/com/android/volley/RequestQueue.java
+++ b/src/main/java/com/android/volley/RequestQueue.java
@@ -38,9 +38,27 @@ public class RequestQueue {
     // TODO: This should not be a generic class, because the request type can't be determined at
     // compile time, so all calls to onRequestFinished are unsafe. However, changing this would be
     // an API-breaking change. See also: https://github.com/google/volley/pull/109
+    @Deprecated // Use RequestEventListener instead.
     public interface RequestFinishedListener<T> {
         /** Called when a request has finished processing. */
         void onRequestFinished(Request<T> request);
+    }
+
+    /** Request event types listeners will be notified about. */
+    public enum RequestEvent {
+        REQUEST_QUEUED,
+        REQUEST_CACHE_LOOKUP_STARTED,
+        REQUEST_CACHE_LOOKUP_FINISHED,
+        REQUEST_PROCESSING_STARTED,
+        REQUEST_PROCESSING_FINISHED,
+        REQUEST_FINISHED
+    }
+
+    /** Callback interface for request life cycle events. */
+    public interface RequestEventListener {
+        // Listener can be called from different threads but in synchronized way. Also it's
+        // guaranteed that the events will be sent in the order they have happened.
+        void onRequestEvent(Request request, RequestEvent event);
     }
 
     /** Used for generating monotonically-increasing sequence numbers for requests. */
@@ -77,6 +95,9 @@ public class RequestQueue {
     private CacheDispatcher mCacheDispatcher;
 
     private final List<RequestFinishedListener> mFinishedListeners = new ArrayList<>();
+
+    /** Collection of listeners for request life cycle events. */
+    private final List<RequestEventListener> mEventListeners = new ArrayList<>();
 
     /**
      * Creates the worker pool. Processing will not begin until {@link #start()} is called.
@@ -217,9 +238,11 @@ public class RequestQueue {
         // If the request is uncacheable, skip the cache queue and go straight to the network.
         if (!request.shouldCache()) {
             mNetworkQueue.add(request);
+            sendRequestEvent(request, RequestEvent.REQUEST_QUEUED);
             return request;
         }
         mCacheQueue.add(request);
+        sendRequestEvent(request, RequestEvent.REQUEST_QUEUED);
         return request;
     }
 
@@ -238,8 +261,33 @@ public class RequestQueue {
                 listener.onRequestFinished(request);
             }
         }
+        sendRequestEvent(request, RequestEvent.REQUEST_FINISHED);
     }
 
+    /** Sends a request life cycle event to the listeners. */
+    void sendRequestEvent(Request request, RequestEvent event) {
+        synchronized (mEventListeners) {
+            for (RequestEventListener listener : mEventListeners) {
+                listener.onRequestEvent(request, event);
+            }
+        }
+    }
+
+    /** Add a listener for request life cycle events. */
+    public void addRequestEventListener(RequestEventListener listener) {
+        synchronized (mEventListeners) {
+            mEventListeners.add(listener);
+        }
+    }
+
+    /** Remove a listener for request life cycle events. */
+    public void removeRequestEventListener(RequestEventListener listener) {
+        synchronized (mEventListeners) {
+            mEventListeners.remove(listener);
+        }
+    }
+
+    @Deprecated // Use RequestEventListener instead.
     public <T> void addRequestFinishedListener(RequestFinishedListener<T> listener) {
         synchronized (mFinishedListeners) {
             mFinishedListeners.add(listener);
@@ -247,6 +295,7 @@ public class RequestQueue {
     }
 
     /** Remove a RequestFinishedListener. Has no effect if listener was not previously added. */
+    @Deprecated // Use RequestEventListener instead.
     public <T> void removeRequestFinishedListener(RequestFinishedListener<T> listener) {
         synchronized (mFinishedListeners) {
             mFinishedListeners.remove(listener);
--- a/src/test/java/com/android/volley/CacheDispatcherTest.java
+++ b/src/test/java/com/android/volley/CacheDispatcherTest.java
@@ -18,6 +18,8 @@ package com.android.volley;
 
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.never;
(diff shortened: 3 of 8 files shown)


src/main/java/com/android/volley/Request.java
```diff
@@ -124,6 +125,8 @@
     @GuardedBy("mLock")
     private NetworkRequestCompleteListener mRequestCompleteListener;
 
+    private Map<String, Object> mTags = new HashMap<>();
```
jpd236: While this might be reasonable given that there's already an mTag, I think it would be unclear to clients how these interact with each other given the same naming.

I'm also not as convinced this piece is as general purpose. The existing tag serves the purpose of making it possible to cancel all requests with a given tag. This does nothing except store app-specific metadata, and in cases like this I'd generally be asking client apps to manage this in their own request subclasses instead of adding it to Volley itself. What does this end up being used for, and is there another way to keep track of this externally instead of in the request itself?

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +38,29 @@
...
     // an API-breaking change. See also: https://github.com/google/volley/pull/109
+    @Deprecated // Use RequestEventListener instead.
     public interface RequestFinishedListener<T> {
         /** Called when a request has finished processing. */
         void onRequestFinished(Request<T> request);
     }
 
+    /** Request event types listeners will be notified about. */
+    public enum RequestEvent {
+        REQUEST_QUEUED,
```
jpd236: these should each have Javadocs

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +38,29 @@
...
+        REQUEST_PROCESSING_STARTED,
+        REQUEST_PROCESSING_FINISHED,
+        REQUEST_FINISHED
+    }
+
+    /** Callback interface for request life cycle events. */
+    public interface RequestEventListener {
+        // Listener can be called from different threads but in synchronized way. Also it's
+        // guaranteed that the events will be sent in the order they have happened.
+        void onRequestEvent(Request request, RequestEvent event);
```
jpd236: Request<?> (otherwise this is a Java warning AFAIK)

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +38,29 @@
...
     // compile time, so all calls to onRequestFinished are unsafe. However, changing this would be
     // an API-breaking change. See also: https://github.com/google/volley/pull/109
+    @Deprecated // Use RequestEventListener instead.
     public interface RequestFinishedListener<T> {
         /** Called when a request has finished processing. */
         void onRequestFinished(Request<T> request);
     }
 
+    /** Request event types listeners will be notified about. */
+    public enum RequestEvent {
```
jpd236: I'm on the fence about this, but feel slightly that this should be a set of @IntDef'd constants rather than an enum given the Android team's long-standing recommendation against enums and the fact that we're a commonly-used library and can't control the optimization processes that apps using it might have configured.

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -217,9 +238,11 @@ public boolean apply(Request<?> request) {
         // If the request is uncacheable, skip the cache queue and go straight to the network.
         if (!request.shouldCache()) {
             mNetworkQueue.add(request);
+            sendRequestEvent(request, RequestEvent.REQUEST_QUEUED);
```
jpd236: this feels like it should be done next to request.addMarker("add-to-queue") once.

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -238,15 +261,41 @@ public boolean apply(Request<?> request) {
                 listener.onRequestFinished(request);
             }
         }
+        sendRequestEvent(request, RequestEvent.REQUEST_FINISHED);
+    }
+
+    /** Sends a request life cycle event to the listeners. */
+    void sendRequestEvent(Request request, RequestEvent event) {
```
jpd236: Request<?>

src/test/java/com/android/volley/CacheDispatcherTest.java
```diff
@@ -231,4 +235,39 @@ public void duplicateSoftExpiredCacheHit_successfulRequest() throws Exception {
         verify(mDelivery)
                 .postResponse(any(Request.class), any(Response.class), any(Runnable.class));
     }
+
+    @Test
+    public void processRequestNotifiesListener() throws Exception {
+        final AtomicBoolean listenerWasCalledOnStart = new AtomicBoolean(false);
+        final AtomicBoolean listenerWasCalledOnFinish = new AtomicBoolean(false);
+
+        RequestQueue.RequestEventListener listener =
```
jpd236: WDYT about a generic test helper class which just records the list of events that were received, and then asserting that the set of events (in order) received matches expectations?

At that point I don't even think you need an explicit listener object, but could just use a Mockito mock instead and verify (inOrder) the calls to onRequestEvent.

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +38,29 @@
...
+        REQUEST_CACHE_LOOKUP_STARTED,
+        REQUEST_CACHE_LOOKUP_FINISHED,
+        REQUEST_PROCESSING_STARTED,
+        REQUEST_PROCESSING_FINISHED,
+        REQUEST_FINISHED
+    }
+
+    /** Callback interface for request life cycle events. */
+    public interface RequestEventListener {
+        // Listener can be called from different threads but in synchronized way. Also it's
```
jjoslin: I feel like you should call out that any event processing should be kept to a minimum or moved to a separate thread to prevent the request from blocking. Alternatively, require an Executor be provided with the listener and just post all of the events to it.

src/main/java/com/android/volley/RequestQueue.java
jpd236: I actually think this doesn't make sense because all Requests must be passed to registered event listeners, even if they don't match the expected generic type (which is unknown at runtime). So while it compiles, it means there's an unsafe cast when passing the request to the event listener, and there's not actually any type safety. <?> accurately notes that this is a Request of unknown type, so we can't make any conclusions about it at compile time.

(See also the note above RequestFinishedListener)

jjoslin: Ahh right, I forget it would be for all Requests :/

src/main/java/com/android/volley/Request.java
artikz: The idea behind the tags here is similar to View tags (https://developer.android.com/reference/android/view/View.html#setTag(java.lang.Object), https://developer.android.com/reference/android/view/View.html#setTag(int,%20java.lang.Object)) - I would like to store additional data associated with the request along with the request.

Although it's still possible to do this externally (e.g., with WeakHashMap), having it here greatly simplifies the design.

And unfortunately storing this in a subclass won't help, as in our specific case we want to have this available for all the requests, no matter whether they are extending our subclass or not.

Having said that I don't feel too strongly, because having the listener APIs will still simplify things. I can revert this part for now and we'll see how it works out without it.

artikz: Removed this for now

src/main/java/com/android/volley/RequestQueue.java
artikz: Done

src/main/java/com/android/volley/RequestQueue.java
artikz: Done

src/main/java/com/android/volley/RequestQueue.java
artikz: Done

src/main/java/com/android/volley/RequestQueue.java
artikz: Done

src/test/java/com/android/volley/CacheDispatcherTest.java
artikz: Done

src/main/java/com/android/volley/Request.java
```diff
@@ -176,6 +179,19 @@ public Object getTag() {
         return mTag;
     }
 
+    public Request<?> setTag(String key, Object tag) {
```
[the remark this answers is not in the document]
artikz: Removed the methods

src/main/java/com/android/volley/Request.java
```diff
@@ -176,6 +179,19 @@ public Object getTag() {
...
+        mTags.put(key, tag);
+        return this;
+    }
+
+    public Object getTag(String key) {
+        return mTags.get(key);
+    }
+
+    public Object removeTag(String key) {
```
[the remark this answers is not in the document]
artikz: Removed the methods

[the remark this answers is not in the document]
artikz: Removed the methods

src/main/java/com/android/volley/RequestQueue.java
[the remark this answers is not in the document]
artikz: Done

artikz: Done

src/main/java/com/android/volley/RequestQueue.java
artikz: Done

src/main/java/com/android/volley/Request.java
```diff
@@ -176,6 +179,19 @@ public Object getTag() {
         return mTag;
     }
 
+    public Request<?> setTag(String key, Object tag) {
+        mTags.put(key, tag);
+        return this;
+    }
+
+    public Object getTag(String key) {
```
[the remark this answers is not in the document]
artikz: Removed the methods

artikz:
I think I addressed all the comments

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +41,56 @@
...
+        /** The request was added to the queue. */
+        public static final int REQUEST_QUEUED = 0;
+        /** The cache lookup started for the request. */
+        public static final int REQUEST_CACHE_LOOKUP_STARTED = 1;
+        /**
+         * The cache lookup finished for the request and cached response is delivered or request is
+         * queued for processing.
+         */
+        public static final int REQUEST_CACHE_LOOKUP_FINISHED = 2;
+        /** The processing (actual network access) started for the request. */
```
jpd236: as the fact that the parenthetical is necessary kind of gets at, "PROCESSING" seems unnecessarily vague here. What's happening is that the request is being sent over the network. Perhaps REQUEST_NETWORK_DISPATCH_STARTED/FINISHED would be clearer, and lead to a more natural Javadoc like "Network dispatch started for the request".

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +41,56 @@
     // TODO: This should not be a generic class, because the request type can't be determined at
     // compile time, so all calls to onRequestFinished are unsafe. However, changing this would be
     // an API-breaking change. See also: https://github.com/google/volley/pull/109
+    @Deprecated // Use RequestEventListener instead.
     public interface RequestFinishedListener<T> {
         /** Called when a request has finished processing. */
         void onRequestFinished(Request<T> request);
     }
 
+    /** Request event types listeners will be notified about. */
```
jpd236: nit: reference {@link RequestEventListener} here to be more precise.

src/main/java/com/android/volley/RequestQueue.java
```diff
@@ -38,11 +41,56 @@
...
+        RequestEvent.REQUEST_PROCESSING_FINISHED,
+        RequestEvent.REQUEST_FINISHED
+    })
+    public @interface RequestEvent {
+        /** The request was added to the queue. */
+        public static final int REQUEST_QUEUED = 0;
+        /** The cache lookup started for the request. */
+        public static final int REQUEST_CACHE_LOOKUP_STARTED = 1;
+        /**
+         * The cache lookup finished for the request and cached response is delivered or request is
```
jpd236: nit: "and the cached response is delivered or the request is queued for network dispatching."

src/test/java/com/android/volley/CacheDispatcherTest.java
```diff
@@ -231,4 +234,22 @@ public void duplicateSoftExpiredCacheHit_successfulRequest() throws Exception {
...
+        RequestQueue.RequestEventListener listener = mock(RequestQueue.RequestEventListener.class);
+        RequestQueue queue = new RequestQueue(mCache, mNetwork, 0, mDelivery);
+        queue.addRequestEventListener(listener);
+        mRequest.setRequestQueue(queue);
+
+        Cache.Entry entry = CacheTestUtils.makeRandomCacheEntry(null, false, false);
+        when(mCache.get(anyString())).thenReturn(entry);
+        mDispatcher.processRequest(mRequest);
+
+        verify(listener)
```
jpd236: I don't think these verify calls are validating the order. See:

https://static.javadoc.io/org.mockito/mockito-core/2.23.0/org/mockito/Mockito.html#in_order_verification

src/main/java/com/android/volley/RequestQueue.java
jjoslin: I haven't had a free minute to review this today and I'll be at an offsite for most of tomorrow. I'll try to get to it in the morning but it may be Monday before I can give it a full review. I hope that's okay.

PR #242, opened 2018-11-28
Clearing NetworkImageView's default image should always be safe.

aaijazi:
This allows callers to pessimistically clear the default image before
setting a new one, if they're unsure which one is set. This is useful
for things like RecyclerViews.


aaijazi:
On second thought, it seems better to just prefer the most recent default and have each setter clear the other default, so I'll put together a PR for that instead.

PR #248, opened 2018-12-13
Update build

elevenfive:
- Update Android Gradle Plugin to 3.2.1
- Use 28.x build tools


src/main/AndroidManifest.xml
```diff
@@ -4,8 +4,6 @@
     android:versionCode="1"
     android:versionName="1.0" >
 
-    <uses-sdk android:minSdkVersion="8" />
```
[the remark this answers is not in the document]
elevenfive: No problem, didn't know about the alternate build systems.

build.gradle
```diff
@@ -47,8 +48,13 @@ group = 'com.android.volley'
 version = '1.2.0-SNAPSHOT'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion = '25.0.2'
+    compileSdkVersion 28
+    buildToolsVersion = '28.0.3'
+
+    defaultConfig {
+        minSdkVersion 8
```
jpd236: is it necessary to specify this here since it's already in the AndroidManifest after your latest commit?

it's minor, but better if it's only specified in one place, or else we might update one in the future and be confused depending on which one takes precedence.

elevenfive: Gotcha.  I switched it to a build variable per https://developer.android.com/studio/build/manifest-build-variables .  I'm not sure how I can test the alternate build system to make sure it works there, any thoughts?

jpd236: Is it not working if you just leave the minSdkVersion out of build.gradle entirely and have it hard-coded in the manifest, like how we had it before? Or does the new gradle version make this a required field for some reason?

elevenfive: New gradle plugin requires it, else it assumes 1

jpd236: Ah okay, that's what I was missing.

In that case let's have it explicitly set as 8 in both places. Just please add a comment to each along the lines of:

```
defaultConfig {
  // Keep in sync with src/main/AndroidManifest.xml
  minSdkVersion 8
}
```

and

```
<!-- Keep in sync with build.gradle -->
<uses-sdk minSdkVersion="8" />
```

build.gradle
```diff
@@ -47,8 +48,14 @@ group = 'com.android.volley'
...
 android {
-    compileSdkVersion 25
-    buildToolsVersion = '25.0.2'
+    compileSdkVersion 28
+    buildToolsVersion = '28.0.3'
+
+    defaultConfig {
+        // Keep in sync with src/main/AndroidManifest.xml
+        minSdkVersion 8
+        targetSdkVersion 28
```
jpd236: sorry for not noticing this before - since this was unset in previous builds, what was the effective value - 1 as well? If so, I'm hesitant to change this to 28 in case that leads to compatibility issues in apps which are using Volley but currently have older targetSdkVersions.

elevenfive: It's unclear what the effective previous target was.  But, it's fine for libraries to have a higher target than the app.  The effective target is determined by the application's target, not any libraries.

jpd236: I think you can just omit the targetSdkVersion altogether (at least I was able to patch in your CL and build without it). The minSdkVersion is needed to build the test manifest since it fails to infer it from the AndroidManifest.xml, but the targetSdkVersion can remain unset, which seems like it would preferable just in case it has unexpected impact.

elevenfive:
No problem, glad to help! Thanks for reviewing.

PR #249, opened 2018-12-14
Migrate to AndroidX

rurimo:
- upgrade gradle version & compileSdkVersion
- migrate to androidx
- refactor dependencies

pre-sized array is not recommended in HotSpot based JVMs.
empty-array is recommended in modern java
```java
Object[] objArray =  collection.toArray(new Object[0]);
```

## reference
- intellij
Call to 'toArray()' with pre-sized array argument 'new PathMatcher[collection.size()]' less... (Ctrl+F1)
There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow.
However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version.
Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array,
if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

- https://code.i-harness.com/en/q/2a80d
- http://dungba.org/arraylist-toarray-myth-debunked/
- https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_new_reflective_array


jpd236:
Just to update you - https://github.com/google/volley/pull/248 was sent out slightly before this which handles the gradle plugin updates, so I'm currently chasing that one first.

Afterwards, I will need to test this internally to make sure that it functions when the code is imported into a few other places (Android open source, Google's internal source tree) - I believe it should, though.

I might also suggest that you split out the pull requests so they're a little more individually friendly to review and manage. This PR makes at least three different and largely independent (AFAICT) changes - gradle updates, migrating to androidx, and avoiding array presizing - and so the review may get bogged down in one area even if the other two are ready to go.

jpd236:
And - thank you for sending this along :)

rurimo:
@jpd236 
Thank you for reply.
I will split out the pull requests. (https://github.com/google/volley/pull/251)
Please close this PR!

PR #251, opened 2018-12-18
Migrate to AndroidX

skydoves:
- migrate to AndroidX
- refactor gradle dependencies and third-party libraries version


jpd236:
already being worked on at https://github.com/google/volley/pull/249 - please make suggestions / comments there if you've handled anything differently

jpd236:
ah, I suppose this is the same user? carry on then :)

skydoves:
Yes, same user. :)
Thank you for your fast reply!

rules.gradle
```diff
@@ -20,15 +20,15 @@ tasks.withType(JavaCompile) {
 }
 
 dependencies {
-  provided "com.android.support:support-annotations:28.0.0"
+    implementation "androidx.annotation:annotation:1.0.1"
 }
 
 // Check if the android plugin version supports unit testing.
-if (configurations.findByName("testCompile")) {
+if (configurations.findByName("testCompile") || configurations.findByName("testImplementation")) {
```
jpd236: since we only apply the dependencies under "testImplementation" and not "testCompile", it doesn't make sense to check for both.

either we want to retain backwards compatibility in this rules.gradle for older gradle plugins, in which case we should continue using compile even if it's deprecated, or we don't care about it any more, in which case we should just use testImplementation and not worry about checking if the android plugin version supports testImplementation, since we're implicitly requiring a new enough version anyway.

there are a lot of places that are pulling in this tree... it's *probably* okay to use testImplementation at this point, but I think it would be safest to just leave this as testCompile instead of testImplementation for now. It's just a gradle warning, right?

skydoves: I agree. It's just a gradle warning like below.
```
Configuration 'testCompile' is obsolete and has been replaced with 'testImplementation' and 'testApi'.
```
So it seems better to continue using compile than implementation.
Thank you for the detail explanation!

jpd236:
Sorry for the delay!

Unfortunately, while the CL works and builds fine, we're running into a number of issues when importing the project into Google's internal source tree due to transitive dependencies on the support library. I've started to work through these, but it may be a while before we're able to get to them all, so I can't yet provide a timeframe on if/when we'll be able to merge this.

This is just cleanup, right? Is it blocking your workflow in some form?

jpd236:
It's been quite a while, but I believe we've found a path forward regarding Google's internal source tree, so I can go ahead and merge this change. Thanks again for your patience!

PR #265, opened 2019-04-02
Fix 'MixedMutabilityReturnType' prone error which lead to compile failed

TangHuaiZhe:
execute `./gradlew build`  and build failed due to Error Prone.

https://errorprone.info/bugpattern/MixedMutabilityReturnType


TangHuaiZhe:
> I signed it!

jpd236:
Thanks and sorry for the breakage. We've been meaning to pin to a specific version of error prone to prevent these kinds of breakages from arising, so I went ahead and sent out https://github.com/google/volley/pull/266 to fix that.

For these specific failures, we document that NetworkResponse#headers and NetworkResponse#allHeaders must not be mutated directly. So rather than make it easier for clients to mutate these fields unexpectedly, I think it would be more consistent with the API and reasonable to always return immutable objects. Since we don't want to depend on a large library like Guava, the best way to do that would likely be to wrap the returned objects in Collections.unmodifiableList/Map.

TangHuaiZhe:
Thank your accurate reply and I'll try  another pull request to wrap the returned objects in Collections.unmodifiableList/Map.

PR #268, opened 2019-04-10
 Allow sending any JSON in JSONArrayRequest

NotSpooky:
Instead of just arrays.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/JsonArrayRequest.java
+++ b/src/main/java/com/android/volley/toolbox/JsonArrayRequest.java
@@ -25,6 +25,7 @@ import com.android.volley.Response.Listener;
 import java.io.UnsupportedEncodingException;
 import org.json.JSONArray;
 import org.json.JSONException;
+import org.json.JSONObject;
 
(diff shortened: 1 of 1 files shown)


NotSpooky:
Already signed the CLA.
Will need to actually clone the project to make unit tests work I guess.

jpd236:
Unfortunately this change would not be backwards compatible in terms of APIs, because JSONArray is not a subclass of JSONObject, so existing callers would break.

You'd need to add a constructor overload instead which takes JSONObject, rather than modifying the existing one.

paulsmithkc:
From a usability standpoint, I've never seen a case where it made sense to send a JSON array as request data over a JSON object.  The problem is that you actually want to send an object and get back an array. But there should be an easy way to implement this without breaking the API.

jpd236:
I'd be happy to review a new PR (or update to this one) which adds a new constructor overload. Not sure what happened to the original author of this one; I didn't mean for my comment to suggest that this was infeasible (hence why it's still open).

paulsmithkc:
New PR created https://github.com/google/volley/pull/406

PR #274, opened 2019-05-30
Allow Volley.newRequestQueue() to be called on main thread

Poligun:
Currently the helper method calls context.getCacheDir() synchronously and causes strict mode violation when invoked on main thread. By using a supplier instead we can have it constructed lazily on the dispatcher's thread.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/Volley.java
+++ b/src/main/java/com/android/volley/toolbox/Volley.java
@@ -86,8 +86,25 @@ public class Volley {
     }
 
     private static RequestQueue newRequestQueue(Context context, Network network) {
-        File cacheDir = new File(context.getCacheDir(), DEFAULT_CACHE_DIR);
-        RequestQueue queue = new RequestQueue(new DiskBasedCache(cacheDir), network);
+        final Context appContext = context.getApplicationContext();
+        // Use a lazy supplier for the cache directory so that newRequestQueue() can be called on
+        // main thread without causing strict mode violation.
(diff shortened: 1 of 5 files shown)


src/main/java/com/android/volley/toolbox/Volley.java
```diff
@@ -86,8 +86,27 @@ public static RequestQueue newRequestQueue(Context context, HttpStack stack) {
     }
 
     private static RequestQueue newRequestQueue(Context context, Network network) {
-        File cacheDir = new File(context.getCacheDir(), DEFAULT_CACHE_DIR);
-        RequestQueue queue = new RequestQueue(new DiskBasedCache(cacheDir), network);
+        final Context appContext = context.getApplicationContext();
+        // Use a lazy supplier for the cache directory so that newRequestQueue() can be called on
+        // main thread without causing strict mode violation.
+        DiskBasedCache.FileSupplier cacheSupplier =
+                new DiskBasedCache.FileSupplier() {
```
jpd236: I don't think the implementation of this FileSupplier needs to support parallel access, since DiskBasedCache serializes access to the directory (or else it would have all sorts of bad races).

I'd probably just suggest you implement it as the simple:

```return new File(appContext.getCacheDir(),  DEFAULT_CACHE_DIR);```

rather than worrying about caching the File instance, since the performance hit should be negligible.

Poligun: Done.

Issue #282, opened 2019-07-17
Ignore Expires header of 0 and -1 for error log


Floens:
When processing the `Expires` header in `HttpHeaderParser`, and when the value is invalid it logs an error.

https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/HttpHeaderParser.java#L138

However, Cloudflare often sets this field to `0` or `-1`, causing the log to be spammed with warnings. I'd rather see the warning not being logged at all, of a simple filter for `0` and `-1`.

jpd236:
This has come up before (#223). As noted then, 0 and -1 are invalid values according to the HTTP RFC, as far as I can tell. But the RFC does say this:

"A cache recipient MUST interpret invalid date formats, especially the value "0", as representing a time in the past (i.e., "already expired")."

which matches Volley's behavior. The only contention here seems to be the logspam about the warnings, and given the frequency of reports about it, I suppose we should just go ahead and accept that these invalid date formats are common enough in practice to not be worth logging warnings about.

jpd236:
A pull request would be welcome here. My proposal would be to drop the message to .v (verbose) for 0 and -1 and keep it at .e otherwise, and we can see if that makes folks happy.

PR #283, opened 2019-07-20
httpheaderparser: log invalid date headers only verbosely when common invalid value.

Floens:
See #282

### DIFF
--- a/src/main/java/com/android/volley/toolbox/HttpHeaderParser.java
+++ b/src/main/java/com/android/volley/toolbox/HttpHeaderParser.java
@@ -135,7 +135,15 @@ public class HttpHeaderParser {
             return newRfc1123Formatter().parse(dateStr).getTime();
         } catch (ParseException e) {
             // Date in invalid format, fallback to 0
-            VolleyLog.e(e, "Unable to parse dateStr: %s, falling back to 0", dateStr);
+            // If the value is either "0" or "-1" we only log to verbose,
+            // these values are pretty common and cause log spam.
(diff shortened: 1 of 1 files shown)


Floens:
I signed the cla

src/main/java/com/android/volley/toolbox/HttpHeaderParser.java
```diff
@@ -135,7 +135,15 @@ public static long parseDateAsEpoch(String dateStr) {
             return newRfc1123Formatter().parse(dateStr).getTime();
         } catch (ParseException e) {
             // Date in invalid format, fallback to 0
-            VolleyLog.e(e, "Unable to parse dateStr: %s, falling back to 0", dateStr);
+            // If the value is either "0" or "-1" we only log to verbose,
+            // these values are pretty common and cause log spam.
+            String message = "Unable to parse dateStr: %s, falling back to 0";
+            if (dateStr.equals("0") || dateStr.equals("-1")) {
```
jpd236: nit: this looks okay in this particular case, but I would prefer:

"0".equals(dateStr) || "1".equals(dateStr)

because this more clearly cannot cause a NullPointerException where one was impossible before. As currently written, you would have to reason about whether parse(dateStr) would have thrown an NullPointerException before this code would have been reached. The answer depends on the implementation of SimpleDateFormat, since that behavior is undocumented, and it looks like it would generally throw a NullPointerException, but better not to rely on that and just gracefully accept a null dateStr here since nothing else is enforcing it clearly.

PR #292, opened 2019-09-07
to  give some info on the first page

jackzhp:


jpd236:
I don't think this is appropriate on the main README given that most apps don't need to worry about this distinction, but I've added an entry to the FAQ here:

https://github.com/google/volley/wiki/Frequently-Asked-Questions#is-volley-asynchronous-or-synchronous

PR #297, opened 2019-10-02
Re-initialize the cache if the directory was deleted.

uhager:
If the user clears the cache from the Android Settings page, the
DiskBasedCache's root directory is deleted. This means that the app will
be running without a cache until it is restarted and the cache
initialized.
This fix initializes the cache when it finds that the root directory no
longer exists.
Note that the first entry after deletion that is put into the cache is
still lost, the cache is only re-initialized when putting that entry
fails. Adding retries would be more complicated, since we would have to
avoid getting into a loop if creating the root directory fails.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
@@ -258,10 +258,11 @@ public class DiskBasedCache implements Cache {
             pruneIfNeeded();
             return;
         } catch (IOException e) {
-        }
-        boolean deleted = file.delete();
-        if (!deleted) {
-            VolleyLog.d("Could not clean up file %s", file.getAbsolutePath());
+            boolean deleted = file.delete();
+            if (!deleted) {
+                VolleyLog.d("Could not clean up file %s", file.getAbsolutePath());
+            }
+            initializeIfRootDirectoryDeleted();
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/android/volley/toolbox/DiskBasedCacheTest.java
```diff
@@ -595,6 +595,21 @@ public void publicMethods() throws Exception {
         assertNotNull(DiskBasedCache.class.getMethod("getFileForKey", String.class));
     }
 
+    @Test
+    public void initializeIfRootDirectoryDeleted() {
+        temporaryFolder.delete();
+        DiskBasedCache uninitializedCache = new DiskBasedCache(temporaryFolder.getRoot(), MAX_SIZE);
```
jpd236: nit - in the spirit of testing more representative behavior, should this unit test perhaps create the cache, call initialize(), and only then delete the temporary folder?

The behavior tested here isn't a supported sequence in general - we don't expect the cache to work if you don't initialize it. It happens to hit the code path you want, but that's not necessarily reliable. On the other hand, we know user devices hit the scenario that an already-initialized cache has the directory deleted from underneath it.

uhager: Done.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -258,10 +258,11 @@ public synchronized void put(String key, Entry entry) {
             pruneIfNeeded();
             return;
```
joebowbeer:
```suggestion
```

If move logic inside the catch block, then remove the return that is no longer needed.

uhager: Done.

jpd236:
Thanks for the change and the review!

PR #308, opened 2019-11-18
Take request's priority into account in WaitingRequestManager

forceLain:
Take request's priority into account in WaitingRequestManager to make sure higher priority requests don't wait for lowe priority requests to be completed.

### DIFF
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -226,16 +226,16 @@ public class CacheDispatcher extends Thread {
                 onNoUsableResponseReceived(request);
                 return;
             }
-            String cacheKey = request.getCacheKey();
+            String waitingKey = getWaitingKey(request);
             List<Request<?>> waitingRequests;
             synchronized (this) {
-                waitingRequests = mWaitingRequests.remove(cacheKey);
+                waitingRequests = mWaitingRequests.remove(waitingKey);
             }
             if (waitingRequests != null) {
                 if (VolleyLog.DEBUG) {
                     VolleyLog.v(
-                            "Releasing %d waiting requests for cacheKey=%s.",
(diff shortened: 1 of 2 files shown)


uhager:
I'm not sure I understand this. If the low priority request is already in the WaitingRequest list, it has already been sent to the network. Now the higher priority request for the same data comes in (assuming same cache key means same URL?), and since there is no waiting request with that priority it is sent to the network right away. But since the low-priority request was sent out first, it will likely also return first (or do you have a network stack that uses priority?); it may already have returned and just be waiting for the lock to send the response to all waiting requests. So you're increasing the network usage, and probably even delaying the high-priority request.
The only case where this helps is if the low-priority request is still on the NetworkDispatcher queue, then the high-priority one gets send first. But you don't know whether that is the case, and either way, you are now slowing down the NetworkDispatcher queue with an additional request.
Am I missing something?

forceLain:
The issue I faced with was: perhaps I submit 10 different low priority requests and all of them get to the NetworkDispatcher queue. Then, when I submit a high priority request with the same URL that the last low priority request has, the higher one gets to the waiting queue and I have to wait for ALL 10 low priority requests to be processed even though I have a higher one.
Changing the waiting key seemed to be an easy fix but now I think maybe it would be better to update the priority of the low priority request in the NetworkDispatcher queue so it gets processed earlier?

uhager:
I don't think changing the priority of a request after it has been added to the queue works. You'd have to remove the current request with low priority and re-add it with the updated priority. This means that WaitingRequestManager would have to keep a list of requests it has already added to the queue, not just the ones that are waiting as it currently does.

jpd236:
Closing this out as it seems like the high-level approach here isn't likely to be beneficial. Feel free to reopen and/or file a bug to discuss further if you think there's something we could be doing better. Thanks for discussing this with us!

PR #315, opened 2020-01-17
Add multipart request

baolanlequang:
- Create a new MutipartRequest file in com.android.volley.toolbox package for multipart request
- Update how to use MutipartRequest in Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,76 @@ importantly, faster.
 
 For more information about Volley and how to use it, visit the [Android developer training
 page](https://developer.android.com/training/volley/index.html).
+
+### How to use MultipartRequest
+
+- MutipartRequest: A mulipart request for multiple files uploading, included text parameters
+- A request example:
+```java
+
+String url = "https://examples.com/multipart_feed";
+MutipartRequest volleyMultipartRequest = new MutipartRequest(Request.Method.POST, url, new Response.Listener<NetworkResponse>() {
+            @Override
+            public void onResponse(NetworkResponse response) {
+                //TODO: handle respone
+            }
+        }, new Response.ErrorListener() {
+            @Override
+            public void onErrorResponse(VolleyError error) {
+                //TODO: handle error
+            }
+        }) {
+            @Override
+            protected Map<String, MultipartData> getByteData() throws AuthFailureError {
+                //add files with filenames
+                Map<String, MultipartData> params = new HashMap<>();
+                try {
+                  byte[] audioData = FileUtils.readFileToByteArray(file);
+                  params.put(fileName, new MultipartData(fileName, audioData, "audio/mpeg"));
+                }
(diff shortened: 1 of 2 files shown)


jpd236:
Thanks for sending this along!

While this is a use case that has been [raised a few times](https://github.com/google/volley/issues?utf8=%E2%9C%93&q=is%3Aissue+multipart), at the moment I don't think this is likely to be used frequently enough to justify including it in Volley itself. My two main arguments for this (beyond usage frequency) would be:

- As I understand it, multipart requests would generally be for uploading files or other content of non-trivial size. This isn't really Volley's intended use case, and it's not something that Volley does particularly well.  See https://groups.google.com/g/volley-users/c/Br7Hme-nbRE/m/b52xaDdMDAAJ for some discussion on this.

- This kind of request seems to be used specifically for HTML form response handling. If you're using Volley and building a native app, I'm not sure what the use case would be for supporting requests issued in this format. I'd generally expect applications to use a simpler structured request/response format, e.g. JSON (with encoded binary data as needed) or protocol buffers etc.

I'm not saying that there aren't potentially reasonable valid use cases here that mostly fit into Volley's mission. But overall, it's very likely that using this type of request will lead to suboptimal performance and is not a pattern we'd want to encourage.

That said, nothing's stopping you from making and using this request class and sharing it outside of Volley itself. It seems like you were able to make this functional without needing changes elsewhere in Volley, which is great. I'd encourage you to share this in its own repository/project. If it proves to be a common/popular use case in terms of issue tracker reports or comments on volley-users, and some use cases are raised to counter my concerns above, then we could consider including it at that time.

Going to close this, but please let me know if you disagree with any of this.

baolanlequang:
@jpd236  The discussion link that you have sent is inaccessible (Error 404).
Multipart for native can you in varies cases, for example:
- A native app has a function that capturing images from its camera and then upload those to its backend with some information (eg: images' names, user information, etc.). We only need to send all of these information in one request instead of upload images first.
This is a case illustrates we need multipart request.
So, please consider it.

jpd236:
Sorry about that! I edited the link - this one should work: https://groups.google.com/d/topic/volley-users/Br7Hme-nbRE/discussion

The use case you mention is *exactly* what Volley is not meant to handle. Pictures from a camera are on the order of megabytes large (or perhaps hundreds of KB if you're doing some sort of compression). As noted in that thread, this is not something that Volley does well. It will store the entire request body in memory, and it does not support streaming, which means that if the network drops out, the user will have to upload the entire batch of images again.

This design choice was intentional to make the API simpler. Volley works well for small RPC-like traffic, but you will almost certainly want to use something else for uploading large files. Nonetheless, if you want to use it this way, you are free to do so, and you can use a class like this one just fine - it's just not something we would want to encourage by including in Volley proper.

PR #317, opened 2020-02-01
Updated README.md

Rec0iL99:
## Proposed changes
Updated README


jpd236:
Whether it's "a" or "an" depends on how you pronounce "H". The more common pronunciation (at least in the US) is "aytch" which makes "an" sound better. See also https://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html.

PR #321, opened 2020-02-24
Send request to network when cache entry parsing fails

uhager:
Also invalidate the corrupted cache entry and set the request's entry to null.

This is probably quite rare, but we should still rather send the request to the network than keep trying to return a potentially corrupted cache entry until it expires.

### DIFF
--- a/src/main/java/com/android/volley/CacheDispatcher.java
+++ b/src/main/java/com/android/volley/CacheDispatcher.java
@@ -159,6 +159,14 @@ public class CacheDispatcher extends Thread {
                             new NetworkResponse(entry.data, entry.responseHeaders));
             request.addMarker("cache-hit-parsed");
 
+            if (!response.isSuccess()) {
+                mCache.invalidate(request.getCacheKey(), true);
+                request.setCacheEntry(null);
+                if (!mWaitingRequestManager.maybeAddToWaitingRequests(request)) {
+                    mNetworkQueue.put(request);
+                }
+                return;
+            }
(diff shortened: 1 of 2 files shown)


src/main/java/com/android/volley/CacheDispatcher.java
```diff
@@ -159,6 +159,15 @@ void processRequest(final Request<?> request) throws InterruptedException {
                             new NetworkResponse(entry.data, entry.responseHeaders));
             request.addMarker("cache-hit-parsed");
 
+            if (!response.isSuccess()) {
```
jpd236: parseNetworkResponse is documented as being allowed to return null in case of an error, which means this would crash if that happens.

That being said, it seems likely that it would already crash in that situation:

- If entry.refreshNeeded(), then it will NPE on line 181 (response.intermediate=true).
- Otherwise, we call mDelivery.postResponse(). From there it depends on the ResponseDelivery implementation, but the default ExecutorDelivery seems like it would crash checking mResponse.isSuccess().

So maybe it's worth filing a separate bug to investigate whether we handle that scenario correctly or should just update the Javadoc to require that callers use Response.error rather than returning null?

uhager:
Given that returning null crashes already, I think it's unlikely that anyone uses a Request that returns null on error. Then it seems better to update the javadoc to reflect the fact that null is already not really supported. But I'll defer to your judgement.

jpd236:
Filed https://github.com/google/volley/issues/322 for the Javadoc.

PR #325, opened 2020-03-26
Add @Nullable annotations to Response

justin-morey:


src/main/java/com/android/volley/Response.java
```diff
@@ -39,7 +41,7 @@
     }
 
     /** Returns a successful response containing the parsed result. */
-    public static <T> Response<T> success(T result, Cache.Entry cacheEntry) {
+    public static <T> Response<T> success(@Nullable T result, @Nullable Cache.Entry cacheEntry) {
```
jpd236: I don't think these were intended to be nullable - certainly not result and probably not cacheEntry as well. The underlying variables are documented as null "in the case of error", which I'd take to mean a call to Response.error() where those fields aren't provided. But in the case of a successful reply, I'd expect you to have a valid result, and probably a valid cache entry as well.

Did you happen to see evidence of nulls being passed in here, or would you be okay leaving these unannotated for now? (Or @NonNull, although I don't think we've historically used that very often given the verbosity).

justin-morey: Volley checks for a null cacheEntry in various places (e.g., NetworkDispatcher.processRequest()), so I believe it should be nullable.

In our project, we use a Response<Void>, passing result=null when calling its success() method, which seems legitimate (although an edge case).  It would certainly be possible for us to pass a non-null object there instead, but Response.result will need to be @Nullable in any case, so this was the natural extension.

jpd236: Definitely fair point for cacheEntry. I do see some other usage of Response.success(null, ...) in cases where the client doesn't care about the response other than that it was successful, so I suppose that's a valid use case as well. 

In light of that, though, I think we should update the Javadoc for result and cacheEntry to match the actual behavior. These are documented as "in the case of error", and I think it would be more accurate to just remove that condition since it can be null in success as well.

justin-morey: Javadoc updated.

Thanks!

PR #330, opened 2020-04-23
Annotate more nullable methods and fields

jbir789:
It will help the IDE and tools better prevent NPEs by knowing what can be null.

### DIFF
--- a/src/main/java/com/android/volley/AuthFailureError.java
+++ b/src/main/java/com/android/volley/AuthFailureError.java
@@ -17,36 +17,39 @@
 package com.android.volley;
 
 import android.content.Intent;
+import androidx.annotation.Nullable;
 
 /** Error indicating that there was an authentication failure when performing a Request. */
 @SuppressWarnings("serial")
 public class AuthFailureError extends VolleyError {
     /** An intent that can be used to resolve this exception. (Brings up the password dialog.) */
-    private Intent mResolutionIntent;
+    @Nullable private Intent mResolutionIntent;
 
     public AuthFailureError() {}
 
-    public AuthFailureError(Intent intent) {
+    public AuthFailureError(@Nullable Intent intent) {
         mResolutionIntent = intent;
     }
 
-    public AuthFailureError(NetworkResponse response) {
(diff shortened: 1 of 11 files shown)


src/main/java/com/android/volley/AuthFailureError.java
```diff
@@ -17,36 +17,40 @@
...
 @SuppressWarnings("serial")
 public class AuthFailureError extends VolleyError {
     /** An intent that can be used to resolve this exception. (Brings up the password dialog.) */
+    @Nullable
     private Intent mResolutionIntent;
 
     public AuthFailureError() {}
 
-    public AuthFailureError(Intent intent) {
+    public AuthFailureError(@Nullable Intent intent) {
```
jpd236, line 31: I'm not sure it makes sense to make these Exception arguments Nullable (for this and other classes). It's true that they're all optional from the perspective of a caller subsequently getting them. But if someone is explicitly calling one of these constructors, I think we want them to be providing non-null arguments. They can always call the no-op empty constructor if they don't want to provide any arguments.

Semantically I think the main difference is that it keeps the door open for us to do a bit more validation on the arguments when you explicitly provide one of the richer objects.

Ultimately it's a minor point, but would it be reasonable to leave the argument annotations they are (except *maybe* the reason argument, since that's explicitly nullable according to the Exception Javadoc itself)? Or do we need this for some reason?

jpd236:
Just noticed that our continuous build was failing but not reporting status to the PR. I think I've triggered a rerun, so please also be sure to address any failures there.

src/main/java/com/android/volley/AuthFailureError.java
jbir789: I agree with you. I was a bit on the fence about doing it for exceptions and ultimately went with it because these arguments can be Nullable, and sometimes even have explicit null checks in the super class. But it is preferable to use the proper constructor when possible. I'll revert those.

jbir789:
Due to manipulation errors, I created a new PR with the latest changes. Closing this one then.

PR #335, opened 2020-05-06
Fix Android Studio project import of Volley.

jpd236:
- Remove minSdkVersion from AndroidManifest.xml since it is already specified
  in build.gradle. Other build systems will need to make similar changes to
  inject the minSdkVersion.

- Use -all instead of -bin for the gradle distribution to prevent AS from
  offering to change this upon first import.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,7 +62,6 @@ android {
     buildToolsVersion = '28.0.3'
 
     defaultConfig {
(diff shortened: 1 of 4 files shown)


jpd236:
@jjoslin I don't seem to be able to request a review for you for some reason - would you mind taking a look?

PR #341, opened 2020-06-04
Fixed testCompile and testApi deprecations

sphill99:
Fixed testCompile and testApi deprecations, by changing to testImplementation. Got rid of warning that compile was obsolete, and has been replaced by implementation.

### DIFF
--- a/rules.gradle
+++ b/rules.gradle
@@ -26,9 +26,9 @@ dependencies {
 // Check if the android plugin version supports unit testing.
 if (configurations.findByName("testCompile")) {
   dependencies {
-      testCompile "junit:junit:4.12"
-      testCompile "org.hamcrest:hamcrest-library:1.3"
-      testCompile "org.mockito:mockito-core:2.19.0"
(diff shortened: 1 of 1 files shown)


rules.gradle
```diff
@@ -26,9 +26,9 @@ dependencies {
 // Check if the android plugin version supports unit testing.
 if (configurations.findByName("testCompile")) {
```
jpd236, line 27: I think this probably needs to be updated as well. Essentially the idea here is to avoid declaring test dependencies if the android plugin version being used doesn't handle them. But if "testCompile" will be removed in the future, this check will start failing.

We can change to "testImplementation" for now. I think we're probably fine to do some broader cleanup here - the fact that we don't just declare these dependencies directly is some legacy where we'd import Volley source code into another build system, and I don't believe there's anyone actively depending on that today. But I don't want to overcomplicate the starter bug :)

sphill99: Thanks! I just made the change in the line you highlighted.

PR #342, opened 2020-06-08
Retry on NoConnectionError

sphill99:
Added the option to retry connecting when receiving a NoConnectionError. Also added a couple tests to make sure this works.

### DIFF
--- a/rules.gradle
+++ b/rules.gradle
@@ -26,9 +26,9 @@ dependencies {
 // Check if the android plugin version supports unit testing.
 if (configurations.findByName("testCompile")) {
   dependencies {
-      testCompile "junit:junit:4.12"
-      testCompile "org.hamcrest:hamcrest-library:1.3"
-      testCompile "org.mockito:mockito-core:2.19.0"
-      testCompile "org.robolectric:robolectric:3.4.2"
+      testImplementation "junit:junit:4.12"
+      testImplementation "org.hamcrest:hamcrest-library:1.3"
+      testImplementation "org.mockito:mockito-core:2.19.0"
+      testImplementation "org.robolectric:robolectric:3.4.2"
   }
 }
--- a/rules.gradle
+++ b/rules.gradle
@@ -24,7 +24,7 @@ dependencies {
 }
 
 // Check if the android plugin version supports unit testing.
-if (configurations.findByName("testCompile")) {
+if (configurations.findByName("testImplementation")) {
   dependencies {
       testImplementation "junit:junit:4.12"
       testImplementation "org.hamcrest:hamcrest-library:1.3"
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -107,6 +107,9 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     /** Whether the request should be retried in the event of an HTTP 5xx (server) error. */
     private boolean mShouldRetryServerErrors = false;
 
+    /** Whether the request should be retried in the event of a NoConnectionError. */
+    private boolean mShouldRetryConnectionErrors = false;
+
     /** The retry policy for this request. */
     private RetryPolicy mRetryPolicy;
 
@@ -531,6 +534,25 @@ public abstract class Request<T> implements Comparable<Request<T>> {
      */
     public final boolean shouldRetryServerErrors() {
         return mShouldRetryServerErrors;
+    }
+
+    /**
+     * Sets whether or not the request should be retried in the event that no connection could be
+     * established.
+     *
+     * @return This Request object to allow for chaining.
+     */
+    public final Request<?> setShouldRetryConnectionErrors(boolean shouldRetryConnection) {
+        mShouldRetryConnectionErrors = shouldRetryConnection;
+        return this;
+    }
+
+    /**
+     * Returns true if this request should be retried in the event that no connection could be
+     * established.
+     */
+    public final boolean shouldRetryConnectionErrors() {
+        return mShouldRetryConnectionErrors;
     }
 
     /**
--- a/src/main/java/com/android/volley/Request.java
+++ b/src/main/java/com/android/volley/Request.java
@@ -107,7 +107,7 @@ public abstract class Request<T> implements Comparable<Request<T>> {
     /** Whether the request should be retried in the event of an HTTP 5xx (server) error. */
     private boolean mShouldRetryServerErrors = false;
 
-    /** Whether the request should be retried in the event of a NoConnectionError. */
+    /** Whether the request should be retried in the event of a {@link NoConnectionError}. */
     private boolean mShouldRetryConnectionErrors = false;
 
     /** The retry policy for this request. */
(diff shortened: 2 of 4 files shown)


rules.gradle
```diff
@@ -24,11 +24,11 @@ dependencies {
 }
```
jbir789, line 24: These changes were made in your previous CL. You might need to merge `google:master` into your branch so it becomes up to date and doesn't show this diff. This page might be useful do do that: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork

src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -178,7 +178,12 @@ public NetworkResponse performRequest(Request<?> request) throws VolleyError {
                 if (httpResponse != null) {
                     statusCode = httpResponse.getStatusCode();
                 } else {
-                    throw new NoConnectionError(e);
+                    if (request.shouldRetryConnectionErrors()) {
+                        attemptRetryOnException("connection", request, new NoConnectionError(e));
```
jbir789, line 182: nit: `"no-connection"` might be more clear

rules.gradle
sphill99: Oh yeah, I was confused why those were still there. Thank you!

src/test/java/com/android/volley/toolbox/BasicNetworkTest.java
```diff
@@ -192,6 +193,43 @@ public void noConnection() throws Exception {
...
+    public void noConnectionRetry() throws Exception {
+        MockHttpStack mockHttpStack = new MockHttpStack();
+        mockHttpStack.setExceptionToThrow(new IOException());
+        BasicNetwork httpNetwork = new BasicNetwork(mockHttpStack);
+        Request<String> request = buildRequest();
+        request.setRetryPolicy(mMockRetryPolicy);
+        request.setShouldRetryConnectionErrors(true);
+        doThrow(new VolleyError()).when(mMockRetryPolicy).retry(any(VolleyError.class));
+        try {
+            httpNetwork.performRequest(request);
```
jbir789, line 206: You could use `Assert#assertThrows`:

```java
assertThrows(VolleyError.class, () -> httpNetwork.performRequest(request))
```

src/test/java/com/android/volley/toolbox/BasicNetworkTest.java
```diff
@@ -192,6 +193,43 @@ public void noConnection() throws Exception {
...
+    public void noConnectionNoRetry() throws Exception {
+        MockHttpStack mockHttpStack = new MockHttpStack();
+        mockHttpStack.setExceptionToThrow(new IOException());
+        BasicNetwork httpNetwork = new BasicNetwork(mockHttpStack);
+        Request<String> request = buildRequest();
+        request.setRetryPolicy(mMockRetryPolicy);
+        request.setShouldRetryConnectionErrors(false);
+        doThrow(new VolleyError()).when(mMockRetryPolicy).retry(any(VolleyError.class));
+        try {
+            httpNetwork.performRequest(request);
```
jbir789, line 225: Same here

src/test/java/com/android/volley/toolbox/BasicNetworkTest.java
```diff
@@ -192,6 +193,43 @@ public void noConnection() throws Exception {
...
 
+    @Test
+    public void noConnectionRetry() throws Exception {
+        MockHttpStack mockHttpStack = new MockHttpStack();
+        mockHttpStack.setExceptionToThrow(new IOException());
+        BasicNetwork httpNetwork = new BasicNetwork(mockHttpStack);
+        Request<String> request = buildRequest();
+        request.setRetryPolicy(mMockRetryPolicy);
+        request.setShouldRetryConnectionErrors(true);
+        doThrow(new VolleyError()).when(mMockRetryPolicy).retry(any(VolleyError.class));
```
jbir789, line 204: Why do we need to throw a VolleyError?

sphill99: That was in the original test to make sure that it did not retry on a NoConnectionError by default. It looks like it's in basically every test for the network. What is the main benefit of using assertThrows instead?

jbir789: assertThrows replaces the entire try / catch, so that's a one line assertion that makes the test clearer, essentially.

This was a different question though, I am not fully sure to understand why we throw an error on retry. Is it not to retry? @jpd236 probably has more context on it.

sphill99: Oh okay, that makes sense. I believe our current tests don't use JUnit5, and assertThrows isn't supported before that.

And yeah, I'm not sure about that.

src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -178,7 +178,12 @@ public NetworkResponse performRequest(Request<?> request) throws VolleyError {
                 if (httpResponse != null) {
                     statusCode = httpResponse.getStatusCode();
                 } else {
-                    throw new NoConnectionError(e);
+                    if (request.shouldRetryConnectionErrors()) {
+                        attemptRetryOnException("no-connection", request, new NoConnectionError(e));
```
jpd236, line 182: nit: looks like Julien might have asked for this, but looking at how we label other connection errors, I think "connection" is probably more appropriate, as in the other cases we log the thing that is failing rather than the full failure (e.g. "server" instead of "server-error", "network" instead of "network-error", etc).

src/test/java/com/android/volley/toolbox/BasicNetworkTest.java
jpd236: The RetryPolicy interface is to have retry throw the provided error if the retry could not be performed:

https://github.com/google/volley/blob/d1a3d5388c79ff1a6fdb904daeff9b39d0bb7d26/src/main/java/com/android/volley/RetryPolicy.java#L55

I'm not certain, but at a glance, I suspect that throwing is simply the easiest way to write this test, because if you don't throw, then Volley will attempt the request a second time. So you'd need to configure the mock chain to handle that case (and, in particular, to not fail the second time or else it would recurse indefinitely). Throwing still ends the request while allowing you to verify that the retry handler was indeed called.

uhager: RetryPolicy throws an error to indicate that no more retries are available, it just returns void to retry. For this test, if it didn't throw at some point, the test would loop with infinite retries.

src/main/java/com/android/volley/Request.java
```diff
@@ -533,6 +536,25 @@ public final boolean shouldRetryServerErrors() {
         return mShouldRetryServerErrors;
     }
 
+    /**
+     * Sets whether or not the request should be retried in the event that no connection could be
+     * established.
+     *
+     * @return This Request object to allow for chaining.
+     */
+    public final Request<?> setShouldRetryConnectionErrors(boolean shouldRetryConnection) {
```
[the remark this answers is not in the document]
sphill99: okay, I just made that fix

src/main/java/com/android/volley/Request.java
```diff
@@ -107,6 +107,9 @@
     /** Whether the request should be retried in the event of an HTTP 5xx (server) error. */
     private boolean mShouldRetryServerErrors = false;
 
+    /** Whether the request should be retried in the event of a NoConnectionError. */
```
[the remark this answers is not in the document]
sphill99: okay, sounds good

src/main/java/com/android/volley/toolbox/BasicNetwork.java
sphill99: sg, I just switched it back

jpd236:
Merged, thanks! For the future, if you add "Fixes \#X" or "Bug \#X" to your commit/PR description, it will automatically link the request to the associated bug, which helps establish a trail. I've gone ahead and done that here.

Issue #181, opened 2018-05-10
Explore the possibility of a new, async network stack


jpd236:
While Volley's client interface is asynchronous, the internals it uses to perform requests are not - there are fixed-size thread pools for cache and network interactions which get blocked for the duration of any requests. This is inherent to the public APIs that Volley currently exposes.

If we wanted to resolve this and support asynchronous HTTP client stacks, we would either need to accept a breaking API change to Volley, or else explore building an alternative network stack API structure that would coexist with the existing synchronous stack. We'd need AsyncRequestQueue/AsyncNetwork/AsyncCache at a minimum.

Notably, there is no method for sending asynchronous HTTP requests built into the Android SDK. (Disk caching might be doable with the nio package). So we wouldn't actually be able to provide an out-of-the-box implementation in Volley's core (at least without taking on a new dependency). We could consider starting separate projects e.g. volley-okhttp or volley-cronet which provide implementations of async queues.

Filing this for tracking ideas, but at the moment this is not on the roadmap.

jjoslin:
I'd be interested in helping out here.

jpd236:
One random piece: per #80, the way our RetryPolicy interface is intended to work is by increasing request timeouts on each subsequent attempt; however, it doesn't introduce any time delay between request attempts. Currently, it's "safe" to throw a sleep() in the retry method to effectively introduce this gap, since it's the NetworkDispatcher thread which is calling it, and that thread is already blocking on network traffic. But in a world with Async dispatchers, we'd want another way to trigger a retry after a delay.

This may involve a similar change as we did for HttpStack -> BaseHttpStack, since RetryPolicy is an interface - deprecate RetryPolicy, introduce a new class which implements it and provide a new default method to return the delay in between requests which defaults to 0.

jpd236:
Very, very rough proof-of-concept here: https://github.com/jpd236/volley/commit/18acc6f3fc1ac85caad6688728cdc366cb9d82ce

There are certainly some details still to be worked out, but I'm reasonably confident that there is a path forward here which would avoid breaking existing clients while still allowing development of a new async layer. The next step here will be to write a design doc.

jpd236:
[...]

PR #344, opened 2020-06-15
Get Implementation in AsyncCache that takes in a callback function

sphill99:
Implemented the get function in AsyncCache that takes in a callback function in two different ways. In the one (DiskBasedAsyncCache), it uses an asynchronous file channel, and for the fallback version, it works very similarly to the current DiskBasedCache. Most of the code in those files are from that, aside from the get method. There are not tests yet for this, as not enough has really been implemented to test it, but was more to check the direction of the code.


jbir789:
Hi Scott, I'd recommend to update the PR title to be a bit more descriptive. I find that the first sentence of your description provides much more context on what this PR is about, so feel free to use it as the title.

Descriptive titles are important is it allows to quickly get a sense of what the PR is about.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,82 @@
...
+    // TODO: Handle callback never being invoked? Match behavior when sync cache throws
+    // RuntimeException.
+
+    @Override
+    public Entry get(String key) {
+        final CountDownLatch latch = new CountDownLatch(1);
+        final AtomicReference<Entry> entryRef = new AtomicReference<>();
+        get(
+                key,
+                new OnGetCompleteCallback() {
```
jbir789, line 23: nit: we could use a lambda expression here for simplicity (I believe [all version of Android can now use a lambda](https://developer.android.com/studio/write/java8-support) when using a recent Android Gradle plugin)

This would convert your code into:

```java
get(
    key, 
    entry -> {
      entryRef.set(entry);
      latch.countDown();
    }
);
```

For context, lambdas are like a more concise way to write an anonymous class. Feel free to take a look at the official java doc about lambdas if it's the first time you hear about them / use them: https://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,82 @@
...
+    public interface OnGetCompleteCallback {
+        void onGetComplete(Entry entry);
+    }
+
+    public abstract void get(String key, OnGetCompleteCallback callback);
+
+    // TODO: Handle callback never being invoked? Match behavior when sync cache throws
+    // RuntimeException.
+
+    @Override
```
jbir789, line 17: Could you also annotate the method as `@Nullable`? It acts a documentation to tell the caller that the return value could be null.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,82 @@
+package com.android.volley;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AsyncCache implements Cache {
+
+    public interface OnGetCompleteCallback {
+        void onGetComplete(Entry entry);
```
jbir789, line 9: Since `entry` can be null, please annotate the parameter as `@Nullable`, so whoever implements the class will be aware that the method parameter can be null.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public abstract void put(String key, Entry entry, OnPutCompleteCallback callback);
+
+    @Override
+    public void put(String key, Entry entry) {
+        final CountDownLatch latch = new CountDownLatch(1);
+        put(
+                key,
+                entry,
+                new OnPutCompleteCallback() {
```
jbir789, line 53: You can use a lambda here as well.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            latch.await();
+            return entryRef.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return null;
+        }
+    }
+
+    // TODO: Implement the rest.
```
jbir789, line 39: nit: did you mean to move this below `put`?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+@SuppressLint("NewApi")
+public class DiskBasedAsyncCache extends AsyncCache {
```
jbir789, line 20: As for `AsyncCache`, make sure to add javadoc to describe what the class does and how to use it.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+                        return rootDirectory;
+                    }
+                };
+        mMaxCacheSizeInBytes = maxCacheSizeInBytes;
+    }
+
+    /** Clears the cache. Deletes all cached files from disk. */
+    @Override
+    public synchronized void clear() {
+        // TODO (sphill99): Implement
```
jbir789, line 66: nit: remove the space between `TODO` and `(sphil99)`

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /** Returns the cache entry with the specified key if it exists, null otherwise. */
+    @Override
+    public void get(String key, OnGetCompleteCallback callback) {
+        final OnGetCompleteCallback cb = callback;
```
jbir789, line 72: nit: prefer spelling out variable names so it's more clear to the reader. You could call it `getCallback`?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+        final DiskBasedAsyncCache.CacheHeader entry = mEntries.get(key);
+        File file = getFileForKey(key);
+        Path path = Paths.get(file.getPath());
+        try {
+            AsynchronousFileChannel afc =
```
jbir789, line 77: This is a closable resource, which means it needs to be closed after being used.

A nice way to achieve this is to use a [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html):

```java
try (AsynchronousFileChannel fileChannel = AsynchronousFileChannel.open(path, StandardOpenOption.READ) {
  // do your work here
} catch (Exception e) {
  // handle exception
}
```

The try-with-resources will automatically close the resource when it completes or when an exception is thrown.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+                                cb.onGetComplete(entry.toCacheEntry(data));
+                            }
+                        }
+
+                        @Override
+                        public void failed(Throwable exc, ByteBuffer attachment) {
+                            cb.onGetComplete(null);
+                        }
+                    });
+        } catch (Exception e) {
```
jbir789, line 99: The `DiskBasedCache` only catchtes `IOException`s. Is there a reason why we want to catch any `Exception`s in here?

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,82 @@
...
+
+public abstract class AsyncCache implements Cache {
+
+    public interface OnGetCompleteCallback {
+        void onGetComplete(Entry entry);
+    }
+
+    public abstract void get(String key, OnGetCompleteCallback callback);
+
+    // TODO: Handle callback never being invoked? Match behavior when sync cache throws
```
sophiahaim, line 14: nit: Add the issue in parentheses:
// TODO(#181): Handle...

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /** Represents a supplier for {@link File}s. */
+    public interface FileSupplier {
+        File get();
+    }
+
+    /** Handles holding onto the cache headers for an entry. */
+    @VisibleForTesting
+    static class CacheHeader {
```
jbir789, line 160: Could `CacheHeader` be shared between this class and `DiskBaseCache`? The implementations don't look the same at this time, but just wondering what's the long term plan and if it would make sense to extract the class to a top-level class so it can be shared.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
```diff
@@ -0,0 +1,454 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DiskBasedAsyncCacheFallback extends AsyncCache {
```
jbir789, line 24: Could you also add doc here to explain how the class differs from `DiskBasedAsyncCache`?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
```diff
@@ -0,0 +1,454 @@
...
+    public synchronized void get(String key, OnGetCompleteCallback callback) {
+        DiskBasedAsyncCacheFallback.CacheHeader entry = mEntries.get(key);
+        // if the entry does not exist, return.
+        if (entry == null) {
+            callback.onGetComplete(null);
+        }
+        File file = getFileForKey(key);
+
+        try {
+            DiskBasedAsyncCacheFallback.CountingInputStream cis =
```
jbir789, line 84: I believe you can use a try-with-resources here as well.

jbir789, line 84: nit: I think you may not need to add the `DiskBaseAsyncCacheFallback` prefix since it's an inner class of this class. (same below)

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+    @Override
+    public synchronized void clear() {
+        // TODO (sphill99): Implement
+    }
+
+    /** Returns the cache entry with the specified key if it exists, null otherwise. */
+    @Override
+    public void get(String key, OnGetCompleteCallback callback) {
+        final OnGetCompleteCallback cb = callback;
+        final DiskBasedAsyncCache.CacheHeader entry = mEntries.get(key);
```
sophiahaim, line 73: Check that entry is non-null

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+        File file = getFileForKey(key);
+        Path path = Paths.get(file.getPath());
+        try {
+            AsynchronousFileChannel afc =
+                    AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+            ByteBuffer buffer = ByteBuffer.allocate((int) file.length());
+            afc.read(
+                    buffer,
+                    0,
+                    buffer,
```
sophiahaim, line 83: Add args to these two to clarify what they are:
/* position= */ 0,
/* attachment= */ buffer

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Oh no, I think I just put that there when I first wrote the try/catch, and never changed it. I'll fix that now.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,249 @@
...
+        // TODO (sphill99): Implement
+    }
+
+    /** Removes the specified key from the cache if it exists. */
+    @Override
+    public synchronized void remove(String key) {
+        // TODO (sphill99): Implement
+    }
+
+    /**
```
sophiahaim, line 135: I think some of these common functions that are the same between your new Cache implementations and the existing one could be factored out into a Util that all three classes could call

You may be able to do the same with some of the CONSTANTS at the top of the file that seem to be the same for all three

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Yeah, I was thinking about making another abstract class that these two extend. This would allow us to not repeat the shared methods/class. Currently, I think the write and read methods would be implemented differently, so I don't know how that'd work. But I think that's definitely better to try to do that.

joebowbeer:
What is test coverage for this new code? I don't see any added tests.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
sphill99: It says that it requires an API of 19 or higher to do a try-with-resources.

sphill99: yep, just fixed this.

src/main/java/com/android/volley/AsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/AsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/AsyncCache.java
sphill99: When I tried to make that change, it says lambda expressions aren't allowed in Java 7. Even though I thought Volley uses java8. I'll look into this.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: added those

sphill99:
@joebowbeer I still haven't written tests for this function. I'm going to write tests for the AsyncCache as a whole, including this implementation. I just wanted to get the general structure reviewed before I move further with the implementations. Currently these implementations aren't being used.

I'll make sure to keep the existing coverage.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Yeah, I want to extract some of the stuff that's repeated out to a higher level class. Should I do that for this PR?

src/main/java/com/android/volley/AsyncCache.java
jpd236: Volley actually is built with minSdkVersion = 8, which is actually Java 6 :)

Lambdas have been backported in principle per https://developer.android.com/studio/write/java8-support - not sure offhand how far back this goes. But I'm not 100% sure what the story is offhand in terms of exposing a library that relies on this support.

While it's not ideal, unless we can pretty quickly and confidently find an answer suggesting we can safely use lambdas without introducing major complexity for apps depending on the library using this desugaring support, we should probably avoid them except when writing code that is guaranteed to run on the right minSdkVersion.

Note that this also includes try-with-resources.

jpd236: FWIW, I'm open to considering increasing the minSdkVersion somewhat. At this point it's exceedingly rare to target anything under API 14, which is what the Android support library requires; that seems like a reasonable point to increase it, if that'd be helpful. But it'd be nice to have a clear benefit to doing so, and 14 won't get us native try-with-resources or lambdas AFAIK.

joebowbeer: To be clear, the current Java target will support try-with-resources if sdk >= 19 

```
compileOptions {
    sourceCompatibility JavaVersion.VERSION_1_7
    targetCompatibility JavaVersion.VERSION_1_7
}
```

Though I think try-with-resources can obscure what's going on, esp. in nested and other complex situations, so no loss of any significance.

https://stackoverflow.com/a/19470405/901597

Java 8 is needed for lambda.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,145 @@
...
+            return;
+        }
+        File file = getFileForKey(key);
+        Path path = Paths.get(file.getPath());
+        try {
+            AsynchronousFileChannel afc =
+                    AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+            ByteBuffer buffer = ByteBuffer.allocate((int) file.length());
+            afc.read(
+                    /* destination */ buffer,
```
sophiahaim, line 87: nit: add "="
e.g. /* destination= */ buffer

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
```diff
@@ -0,0 +1,152 @@
...
+    /** Total amount of space currently used by the cache in bytes. */
+    private long mTotalSize = 0;
+
+    /** The supplier for the root directory to use for the cache. */
+    private final DiskBasedCacheUtility.FileSupplier mRootDirectorySupplier;
+
+    /** The maximum size of the cache in bytes. */
+    private final int mMaxCacheSizeInBytes;
+
+    /** Default maximum disk usage in bytes. */
```
sophiahaim, line 35: Are these used anywhere yet? If not, I think hold off on including them until they're used.

When you do need to add them, I think you can put them in your new Utility class since they are the same for both implementations (async and fallback)

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
```diff
@@ -0,0 +1,152 @@
...
+
+    /** Clears the cache. Deletes all cached files from disk. */
+    @Override
+    public synchronized void clear() {
+        // TODO (sphill99): Implement
+    }
+
+    /** Returns the cache entry with the specified key if it exists, null otherwise. */
+    @Override
+    public synchronized void get(String key, OnGetCompleteCallback callback) {
```
sophiahaim, line 68: This is really similar to the existing DiskBasedCache, except instead of returning null or entry.toCacheEntry, you call callback.onGetComplete(null/entry.toCacheEntry)

I'm wondering if it would be possible to use the existing DiskBasedCache in here and just call #get, then do callback.onGetComplete with whatever it returns. There may be a couple ways of approaching this, so happy to discuss more offline too.

I think it would generally simplify this class if you can leverage the existing implementation.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @param key Cache key
+     * @param callback Callback function that sets the retrieved entry and lets the program know it
+     *     is finished retrieving information.
+     */
+    public abstract void get(String key, OnGetCompleteCallback callback);
+
+    @Nullable
+    @Override
+    public Entry get(String key) {
```
jpd236, line 32: I think we should mark the legacy methods as final - subclasses of AsyncClass should not be modifying these implementations, only implementing the async versions. (At least until someone can provide a valid use case for it; until then, the tighter we keep the API, the stronger the guarantees the implementation can offer).

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,96 @@
...
+                    @Override
+                    public void onGetComplete(Entry entry) {
+                        entryRef.set(entry);
+                        latch.countDown();
+                    }
+                });
+        try {
+            latch.await();
+            return entryRef.get();
+        } catch (InterruptedException e) {
```
jpd236, line 47: We should probably log the exception here as well (with VolleyLog). Otherwise, the operation might not have gone through, but apart from the thread's interrupt status (which is often unchecked by callers), there'd be no visible indication that something had been canceled.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
jpd236: +1 - for the fallback, we should try to share as much code with possible as the existing implementation rather than duplicating it.

In fact, though, I wonder if we really want to have a second class at all. The reality is that it's confusing to have a class extend AsyncCache but not actually be Async - how should the user of this object determine whether the cache supports true Async operation (and thus doesn't need to be called on a blocking thread) or not (and thus does)?

The class which uses this will ultimately need to make a decision based on the capabilities of the cache to either use the fully async API or else send the task to a blocking thread pool. But in the latter case, there's actually no reason to use the async API methods, since they're blocking anyway.

This might become clearer when we actually implement the async cache handling step, but my instinct here is that we actually don't need this class, and should instead have the dispatcher accept either a vanilla Cache (which defaults to DiskBasedCache) or an AsyncCache (which defaults to DiskBasedAsyncCache, where supported). If the cache is an instance of AsyncCache, it knows it can safely use the async methods. If not, it should use the fallback and schedule the read for a blocking background thread.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,145 @@
...
+    private long mTotalSize = 0;
+
+    /** The supplier for the root directory to use for the cache. */
+    private final DiskBasedCacheUtility.FileSupplier mRootDirectorySupplier;
+
+    /** The maximum size of the cache in bytes. */
+    private final int mMaxCacheSizeInBytes;
+
+    /** Default maximum disk usage in bytes. */
+    private static final int DEFAULT_DISK_USAGE_BYTES = 5 * 1024 * 1024;
```
jpd236, line 37: can we share magic constants like this with DiskBasedCache at least until we need to fork them? e.g. by putting them in DiskBasedCacheUtility

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,145 @@
...
+            ByteBuffer buffer = ByteBuffer.allocate((int) file.length());
+            afc.read(
+                    /* destination */ buffer,
+                    /* position */ 0,
+                    /* attachment */ buffer,
+                    new CompletionHandler<Integer, ByteBuffer>() {
+                        @Override
+                        public void completed(Integer result, ByteBuffer attachment) {
+                            if (attachment.hasArray()) {
+                                int offset = attachment.arrayOffset();
```
jpd236, line 94: nit: looks like this variable is unused

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,145 @@
...
+            AsynchronousFileChannel afc =
+                    AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+            ByteBuffer buffer = ByteBuffer.allocate((int) file.length());
+            afc.read(
+                    /* destination */ buffer,
+                    /* position */ 0,
+                    /* attachment */ buffer,
+                    new CompletionHandler<Integer, ByteBuffer>() {
+                        @Override
+                        public void completed(Integer result, ByteBuffer attachment) {
```
jpd236, line 92: wondering whether for safety, we need to read the value of result and truncate the entry to that size (/ handle if it is -1, per the Javadoc).

With the old cache, we'd occasionally see strange bugs due to files being corrupted, possibly due to parallel access. While that's a misuse of Volley if so, it is good if we can gracefully fall back to the extent possible. And it's possible the size changes between when we read the length / allocate the buffer and when we complete the actual data read. So the buffer might have garbage data at the end (if the file was truncated), or might be incomplete (if it grew). In both cases we should probably assume the entry is invalid and avoid returning it.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,145 @@
...
+                            if (attachment.hasArray()) {
+                                int offset = attachment.arrayOffset();
+                                byte[] data = attachment.array();
+                                cb.onGetComplete(entry.toCacheEntry(data));
+                            }
+                        }
+
+                        @Override
+                        public void failed(Throwable exc, ByteBuffer attachment) {
```
jpd236, line 101: log the exception so the failure is visible

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,145 @@
...
+                            cb.onGetComplete(null);
+                        }
+                    });
+        } catch (IOException e) {
+            cb.onGetComplete(null);
```
jpd236, line 106: log exception

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I added a check to see if the file length is the same as the size allocated before the read.

src/main/java/com/android/volley/AsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
sphill99: I changed it so the fallback now doesn't repeat the code. But yeah, that makes sense.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
sphill99: I removed these for now

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
sphill99: added

sphill99:
I should have made appropriate edits for all the PR comments, or left questions on them. It is ready to be reviewed again.

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -0,0 +1,147 @@
...
+/** Handles holding onto the cache headers for an entry. */
+@VisibleForTesting
+class CacheHeader {
+    /**
+     * The size of the data identified by this CacheHeader on disk (both header and data).
+     *
+     * <p>Must be set by the caller after it has been calculated.
+     *
+     * <p>This is not serialized to disk.
+     */
```
jbir789, line 21: This comment looks intended for `long size;` below

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -0,0 +1,147 @@
...
+    /** Magic number for current version of cache file format. */
+    private static final int CACHE_MAGIC = 0x20150306;
+
+    long size;
+
+    /** The key that identifies the cache entry. */
+    final String key;
+
+    /** ETag for cache coherence. */
+    final String etag;
```
jbir789, line 31: nit: please annotate as `@Nullable`

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jbir789: Since we can't use try-with-resources. make sure to close the file channel when the try/catch block is completed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
jpd236: That's certainly better, but then why would we need it at all per my comment? Would it make sense to hold off on adding it unless/until we see how it would be helpful/needed, or do you see a reason that the next level up would make use of this, given that it will need to know how to dispatch the cache operations?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,136 @@
...
+            afc.read(
+                    /* destination= */ buffer,
+                    /* position= */ 0,
+                    /* attachment= */ buffer,
+                    new CompletionHandler<Integer, ByteBuffer>() {
+                        @Override
+                        public void completed(Integer result, ByteBuffer attachment) {
+                            // if the file size changes, return null
+                            if (size != file.length()) {
+                                cb.onGetComplete(null);
```
jpd236, line 80: maybe log a warning in this case since it implies the file changed while we were reading it.

src/main/java/com/android/volley/toolbox/CacheHeader.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: added

joebowbeer: Can this mean that the read terminated early? I would treat this as a failure.

src/main/java/com/android/volley/AsyncCache.java
joebowbeer: Consider awaitUninteruptibly. Is this thread exposed enough to be interrupted? If not, this handling is just another section of untested code.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I think it means the file changed between the time we opened the channel and finished the reading. The size allocated and the file length shouldn't be affected by the success of the read. I think if the read fails, it would never reach the completed block.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,148 @@
...
+        } catch (IOException e) {
+            VolleyLog.d("%s: %s", file.getAbsolutePath(), e.toString());
+            cb.onGetComplete(null);
+        } finally {
+            if (afc != null) {
+                try {
+                    afc.close();
+                } catch (IOException e) {
+                    VolleyLog.d("%s: %s", file.getAbsolutePath(), e.toString());
+                    cb.onGetComplete(null);
```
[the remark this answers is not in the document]
sphill99: oh thanks, just fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCacheFallback.java
sphill99: oh okay, I understand what you're saying now. I just deleted it in the last commit.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
joebowbeer: In either case, it sounds like an error to me. I don't think it serves much purpose to check the file length if it will be ignored for all practical purposes. I would either ignore the file length and use the bytes read, if this is reliable, or use the file length as a correctness check, if not.

sphill99:
I think I addressed all PR comments. It is ready to be reviewed again.

src/main/java/com/android/volley/AsyncCache.java
jpd236: awaitInterruptibly isn't a method on CountDownLatch.

While I agree interrupt handling is a mess in Java, I think we can at least ensure we handle them correctly to the extent possible within Volley (we do have RequestQueue#stop which would stop the executors if we control them).

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public final void put(String key, Entry entry) {
+        final CountDownLatch latch = new CountDownLatch(1);
+        put(
+                key,
+                entry,
+                new OnPutCompleteCallback() {
+                    @Override
+                    public void onPutComplete(boolean success) {
+                        if (success) {
```
jpd236, line 71: AFAICT we need to count down the latch regardless of whether the put succeeds or fails. If we don't, the caller will just be blocking indefinitely even though nothing is still happening.

put doesn't provide a way to return an error, so we should make sure it's logged somewhere in the stack, but otherwise we should just return control to the caller once the put is done.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,147 @@
...
+import java.nio.channels.CompletionHandler;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * AsyncCache implementation that uses Java NIO's AsynchronousFileChannel to perform asynchronous
+ * disk reads and writes. This should only be used by devices with an API level of 26 or above.
```
jpd236, line 19: I think you can remove this sentence and the `@SuppressLint` annotation and just add a `@RequiresApi(Build.VERSION_CODES.O)` instead. This is preferable because:

-  Instead of suppressing all lint checks, it only disables them for APIs added up through O. If you make an API call to something added in P, it will still fail, which is good as we would need additional checks.
-  It will enforce that the caller checks for O+ before using the class (or that it has a RequiresApi annotation of its own)
-  Its self-documenting - no need to repeat it in the Javadoc.

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -0,0 +1,148 @@
...
+import com.android.volley.Cache;
+import com.android.volley.Header;
+import com.android.volley.VolleyLog;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Collections;
+import java.util.List;
+
+/** Handles holding onto the cache headers for an entry. */
+@VisibleForTesting
```
jpd236, line 14: this annotation doesn't make sense here; this class needs to be package-visible in order for other classes in this package to use it.

We use this annotation when we increase the visibility beyond what it otherwise would be for the _sole_ purpose of making it accessible in a unit test.

You should see Android Studio complaining about this, e.g. where DiskBasedAsyncCache is calling:

`cb.onGetComplete(entry.toCacheEntry(data));`

because it thinks toCacheEntry is only visible for testing, but here we're using it from prod code.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,147 @@
...
+                afc.read(
+                        /* destination= */ buffer,
+                        /* position= */ 0,
+                        /* attachment= */ buffer,
+                        new CompletionHandler<Integer, ByteBuffer>() {
+                            @Override
+                            public void completed(Integer result, ByteBuffer attachment) {
+                                // if the file size changes, return null
+                                if (size != file.length()) {
+                                    VolleyLog.d(
```
jpd236, line 81: logging stuff:

1) use e instead of d - this is an error, not debugging info
2) You can do `VolleyLog.e("File changed while reading: %s", file.getAbsolutePath())` - no need to use format args for a constant string. Also, I think it's clearer to have the error be the first thing, and specific details like the file name come second.
3) Here and throughout, it should be "%s", not "s%", no?

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,97 @@
...
+                            latch.countDown();
+                        } else {
+                            return;
```
jpd236, line 74: no need for "return" statement as the last statement in a void method, here and below.

AS highlights these in yellow for me - do you not see that too? In general, we should aim to have no warnings in new code.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,147 @@
...
+                                            "s% s%",
+                                            file.getAbsolutePath(), "file changed while reading");
+                                    cb.onGetComplete(null);
+                                    return;
+                                }
+                                if (attachment.hasArray()) {
+                                    final int offset = attachment.arrayOffset();
```
jpd236, line 88: what's the exact error message you get when you remove this line?

I think it's probably an ErrorProne thing which should probably be suppressed by adding a suppression for that, not by introducing a new variable that is itself unused and a compiler warning. arrayOffset() does nothing internally; we know there is a backing array because we allocated with ByteBuffer.allocate() above.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,147 @@
...
+                                    byte[] data = attachment.array();
+                                    cb.onGetComplete(entry.toCacheEntry(data));
+                                }
+                            }
+
+                            @Override
+                            public void failed(Throwable exc, ByteBuffer attachment) {
+                                VolleyLog.d("%s: %s", file.getAbsolutePath(), exc.toString());
```
jpd236, line 96: read failures are abnormal - use e, and log the full stack trace:

VolleyLog.e(exc, "Failed to read file %s", file.getAbsolutePath());

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jpd236: better, but see above note regarding how to handle errors

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jpd236: I think the confusion here is - this should say "if (size != result)". We shouldn't compare the file size before and after - we should compare the size of the file when we allocated the buffer to the number of bytes we actually read.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
joebowbeerxealth: Line 107 is still potentially a second call to the callback

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: but the number of bytes we actually read is dependent upon the size of the buffer we allocate. This would work in the case of the read bytes being less than the allocated, but it will never be more.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: The error is: "ByteBuffer.array() shouldn't be called unless ByteBuffer.arrayOffset() is used or if the ByteBuffer was initialized using ByteBuffer.wrap() or ByteBuffer.allocate().
                                    byte[] data = attachment.array();
                                                                  ^
    (see https://errorprone.info/bugpattern/ByteBufferBackingArray)"

I believe there is a way to just suppress the warning though. So I'll figure that out.

jpd236: One option would be to just pass Void/null for the attachment parameter and access buffer directly - that would prove to error prone that it's actually the same object which came from ByteBuffer.allocate. As it stands, ErrorProne has no way of knowing this.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jpd236: true. but we should still look at the number of bytes read and make sure it's not less, which could happen even if the file size is now the same.

I think checking for result != size || current file size != size would be reasonable. Or, just the former, and let the latter be handled by integrity checks on the cache header.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Yeah,  I think that should work

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Okay, I'll make that fix

src/main/java/com/android/volley/AsyncCache.java
sphill99: would it be logged in this method, or in the async implementation, where we know what happened?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Fixed this

src/main/java/com/android/volley/toolbox/CacheHeader.java
sphill99: fixed

src/main/java/com/android/volley/AsyncCache.java
joebowbeer: At least a Future (incl. CompletableFuture) can be canceled.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,147 @@
...
+        final CacheHeader entry = mEntries.get(key);
+        // if the entry does not exist, return null.
+        if (entry == null) {
+            cb.onGetComplete(null);
+            return;
+        }
+        final File file = getFileForKey(key);
+        final int size = (int) file.length();
+        Path path = Paths.get(file.getPath());
+        AsynchronousFileChannel afc = null;
```
joebowbeer, line 67: I would move this inside the outer try, which should be a try-catch(IOException), and open a try-finally to close the afc on the next line:

```
try {
  final AsynchronousFileChannel afc = open()
  try {
    // I/O
  } finally {
    afc.close()
  }
} catch (IOException e) {
  // handle e
}
```

See https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/DiskBasedCache.java#L148

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
joebowbeer: Line 107 is still potentially a second invocation of callback.

See comment above and this example of another way to structure this:

https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/DiskBasedCache.java#L148

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Since the class now requires a higher android api, I switched it to use a try-with-resources. I think this should allow it to just be a simple try-catch, and gets rid of these issues.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param callback Callback function that sets the retrieved entry and lets the program know it
+     *     is finished retrieving information.
+     */
+    public abstract void get(String key, OnGetCompleteCallback callback);
+
+    @Nullable
+    @Override
+    public final Entry get(String key) {
+        final CountDownLatch latch = new CountDownLatch(1);
+        final AtomicReference<Entry> entryRef = new AtomicReference<>();
```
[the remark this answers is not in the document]
sphill99: Ended up not having AsyncCache implement Cache at all, so there was no long a need for this.

sphill99:
I've addressed all the comments and concerns. I updated the try-try finally-catch block, into a try-with-resources, now that we are requiring that SDK. I also made AsyncCache not implement Cache anymore.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,42 @@
+package com.android.volley;
+
+import androidx.annotation.Nullable;
+
+/**
+ * AsyncCache is an abstract class which implements the Cache interface for backwards compatibility,
```
jpd236, line 6: nit: update Javadoc now that we no longer implement Cache. My suggestion would just be:

/** Asynchronous equivalent to the {@link Cache} interface. */

As that is a simple summary of what this class is, and links to Cache in a more structured form so you can leverage the existing docs there if someone is looking for more detail.

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,42 @@
...
+ * although it is not meant to be used as a Cache directly. For each method in the Cache interface
+ * (get, put, initialize, invalidate, remove, clear), it defines a corresponding abstract version
+ * which takes a Callback to be invoked upon completion instead of returning the result directly.
+ * The methods in the cache interface are implemented by making calls to the asynchronous method and
+ * blocking until they complete.
+ */
+public abstract class AsyncCache {
+
+    public interface OnGetCompleteCallback {
+        void onGetComplete(@Nullable Cache.Entry entry);
```
jpd236, line 16: nit: Could you add a Javadoc for this method? It doesn't have to be super-detailed, but one thing I think would be good to include is when/why entry would be null, and what we do in the event of an error. Something like:

```java
/**
 * Invoked when the read from the cache is complete.
 *
 * @param entry The entry read from the cache, or null if the cache did not contain an entry for the given key or
 *         if the entry couldn't be read.
 */
```

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,42 @@
...
+     * Retrieves an entry from the cache and send it back through the callback function
+     *
+     * @param key Cache key
+     * @param callback Callback function that sets the retrieved entry and lets the program know it
+     *     is finished retrieving information.
+     */
+    public abstract void get(String key, OnGetCompleteCallback callback);
+
+    public interface OnPutCompleteCallback {
+        void onPutComplete(boolean success);
```
jpd236, line 29: Cache#put returns void today, so the introduction of a success boolean is a new concept that I'm not sure we should introduce. Since the focus here is get, I might suggest leaving this out for now.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,105 @@
...
+                    public File get() {
+                        return rootDirectory;
+                    }
+                };
+    }
+
+    /** Returns the cache entry with the specified key if it exists, null otherwise. */
+    @Override
+    public void get(String key, OnGetCompleteCallback callback) {
+        final OnGetCompleteCallback cb = callback;
```
jpd236, line 53: I think this line of code is only needed to mark this as final so the inner class can use it? If so, you can just mark the callback param as final (this doesn't require changes to AsyncCache) and use that directly.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,105 @@
...
+                        }
+
+                        @Override
+                        public void failed(Throwable exc, Void v) {
+                            VolleyLog.e(exc, "Failed to read file %s", file.getAbsolutePath());
+                            cb.onGetComplete(null);
+                        }
+                    });
+        } catch (IOException e) {
+            VolleyLog.e("%s %s", file.getAbsolutePath(), e.toString());
```
jpd236, line 91: VolleyLog.e(e, "Failed to read file %s", file.getAbsolutePath())

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: yep, I didn't realize I could do that

sphill99:
DiskBasedCache and DiskBasedCache tests now uses the shared classes. I moved some stuff (like CountingInputStream) back from DiskBasedCacheUtility to DiskBasedCache, since the async version does not use it. The tests were also refactored and test the code in its new location.

All comments should have been addressed now.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -0,0 +1,143 @@
...
+class DiskBasedCacheUtility {
+
+    /** Default maximum disk usage in bytes. */
+    static final int DEFAULT_DISK_USAGE_BYTES = 5 * 1024 * 1024;
+
+    /** High water mark percentage for the cache */
+    @VisibleForTesting static final float HYSTERESIS_FACTOR = 0.9f;
+
+    /** Magic number for current version of cache file format. */
+    static final int CACHE_MAGIC = 0x20150306;
```
jbir789, line 24: Note that I don't see this being used outside of CacheHeader, so you could consider removing it from here and keep the one in CacheHeader

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -0,0 +1,98 @@
...
+                        }
+                    });
+        } catch (IOException e) {
+            VolleyLog.e(e, "Failed to read file %s", file.getAbsolutePath());
+            callback.onGetComplete(null);
+        }
+    }
+
+    /** Returns a file object for the given cache key. */
+    public File getFileForKey(String key) {
```
jbir789, line 95: Should this be private?

joebowbeer:
> DiskBasedCache and DiskBasedCache tests now uses the shared classes. I moved some stuff (like CountingInputStream) back from DiskBasedCacheUtility to DiskBasedCache, since the async version does not use it. The tests were also refactored and test the code in its new location.

Thanks!

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: yep, fixed

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: Yeah, that's smart. I'll make that change.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -374,14 +369,21 @@ private void removeEntry(String key) {
         }
     }
 
+    InputStream createInputStream(File file) throws FileNotFoundException {
```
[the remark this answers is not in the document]
sphill99: I had moved them outside of DiskBasedCache, but now since they're there again, I'll add it back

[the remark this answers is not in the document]
jpd236: nit: it would lower the diff output if you moved them back to the same location they were in before, which helps in reviewing

src/test/java/com/android/volley/toolbox/DiskBasedCacheTest.java
```diff
@@ -501,86 +503,87 @@ public void testCountingInputStreamByteCount() throws IOException {
     public void testEmptyReadThrowsEOF() throws IOException {
         ByteArrayInputStream empty = new ByteArrayInputStream(new byte[] {});
         exception.expect(EOFException.class);
-        DiskBasedCache.readInt(empty);
+        DiskBasedCacheUtility.readInt(empty);
```
jpd236, line 506: For tests which solely exercise methods in DiskBasedCacheUtility, we should move their tests to a new DiskBasedCacheUtilityTest rather than having them in here, since they aren't specific to DiskBasedCache any more.

sphill99: Fixed this

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -0,0 +1,27 @@
...
+         *
+         * @param entry The entry read from the cache, or null if the read failed or the key did not
+         *     exist in the cache.
+         */
+        void onGetComplete(@Nullable Cache.Entry entry);
+    }
+
+    /**
+     * Retrieves an entry from the cache and send it back through the callback function
+     *
```
sophiahaim, line 20: nit: "sends it back"

Also would be good to link to the callback – {@link OnGetCompleteCallback#onGetComplete}

src/test/java/com/android/volley/toolbox/DiskBasedCacheTest.java
sophiahaim: Scott, did you upload a new DiskBasedCacheUtilityTest class for the tests you removed from here? I don't see one currently

src/main/java/com/android/volley/AsyncCache.java
sphill99: changed this. I realized I never added the DiskBasedCacheUtilityTest class to the commit, since it was untracked. Should have just fixed that as well.

sphill99:
Uploaded the new test class and fixed the javadocs mistakes.

joebowbeer:
By the way, there are 100+ warnings in the Travis build mostly/all about annotations:

https://travis-ci.org/github/google/volley/builds/701650084

I wish the Travis build (or GitHub Action?) reported code coverage.

jpd236:
Filed https://github.com/google/volley/issues/345 for the warnings - these are unrelated to this PR and not a major problem since they only impact the Javadoc (if at all). Coverage is a nice to have, but I don't see it as super critical since we can watch for it in PRs; the project is not that big and doesn't have many people approving changes.

PR #347, opened 2020-06-29
Put implementation for async cache that uses callback method

sphill99:
This is the put methods implementation for the new async cache. Instead of using a basic output stream, we add the data to a bytebuffer, then use an asynchronous write to the specified file. The write header method works differently from before, as it just puts the information into the buffer, and then writes everything to the file all at once.

There were some non-static methods in DiskBasedCache that I made into static methods and moved them to the utility class. 

In the old read/write methods, I got rid of warnings where we were shifting a value by 0 bits. Can change back if that's preferred.

### DIFF
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -25,7 +25,7 @@ public abstract class AsyncCache {
     public abstract void get(String key, OnGetCompleteCallback callback);
 
     public interface OnPutCompleteCallback {
-        /** Invoked when the put to the cache is complete */
+        /** Invoked when the put to the cache is complete. */
         void onPutComplete();
     }
 
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -29,6 +29,14 @@ public abstract class AsyncCache {
         void onPutComplete();
     }
 
+    /**
+     * Writes a {@link Cache.Entry} to the cache, and calls {@link
+     * OnPutCompleteCallback#onPutComplete} after the operation is finished.
+     *
+     * @param key Cache key
+     * @param entry The entry to be written to the cache
+     * @param callback Callback that will be notified when the information has been written
+     */
     public abstract void put(String key, Cache.Entry entry, OnPutCompleteCallback callback);
 
     // TODO(#181): Implement the rest.
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -25,6 +25,7 @@ public abstract class AsyncCache {
     public abstract void get(String key, OnGetCompleteCallback callback);
 
     public interface OnPutCompleteCallback {
+        /** Invoked when the put to the cache is complete */
         void onPutComplete();
     }
 
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -24,5 +24,11 @@ public abstract class AsyncCache {
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
 
+    public interface OnPutCompleteCallback {
+        void onPutComplete();
+    }
+
+    public abstract void put(String key, Cache.Entry entry, OnPutCompleteCallback callback);
+
     // TODO(#181): Implement the rest.
 }
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -16,6 +16,9 @@ class CacheHeader {
     /** Magic number for current version of cache file format. */
     private static final int CACHE_MAGIC = 0x20150306;
 
+    /** Bits required to write 4 longs and 3 ints */
+    private static final int HEADER_SIZE = 44;
+
     /**
      * The size of the data identified by this CacheHeader on disk (both header and data).
      *
@@ -160,16 +163,16 @@ class CacheHeader {
 
     /** Gets the size of the header in bytes */
     int getHeaderSize() throws IOException {
-        int x = 44; // 4 longs, 3 ints
+        int size = 0;
         try {
-            x += key.getBytes("UTF8").length;
+            size += key.getBytes("UTF8").length;
             if (etag != null) {
-                x += etag.getBytes("UTF8").length;
+                size += etag.getBytes("UTF8").length;
             }
         } catch (UnsupportedEncodingException e) {
-            VolleyLog.e(e.toString());
+            VolleyLog.e(e, "UTF-8 is unsupported");
         }
-        x += DiskBasedCacheUtility.headerListSize(allResponseHeaders);
-        return x;
+        size += DiskBasedCacheUtility.headerListSize(allResponseHeaders);
+        return size + HEADER_SIZE;
     }
 }
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -15,8 +15,8 @@ class CacheHeader {
     /** Magic number for current version of cache file format. */
     private static final int CACHE_MAGIC = 0x20150306;
 
-    /** Bits required to write 4 longs and 3 ints */
-    private static final int HEADER_SIZE = 44;
+    /** Bits required to write 6 longs and 1 int. */
+    private static final int HEADER_SIZE = 52;
 
     /**
      * The size of the data identified by this CacheHeader on disk (both header and data).
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -6,6 +6,8 @@ import com.android.volley.Header;
 import com.android.volley.VolleyLog;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.UnsupportedEncodingException;
+import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.List;
 
@@ -142,5 +144,30 @@ class CacheHeader {
             VolleyLog.d("%s", e.toString());
             return false;
         }
+    }
+
+    void writeHeader(ByteBuffer buffer) throws IOException {
+        buffer.putInt(CACHE_MAGIC);
+        DiskBasedCacheUtility.writeString(buffer, key);
+        DiskBasedCacheUtility.writeString(buffer, etag);
+        buffer.putLong(serverDate);
+        buffer.putLong(lastModified);
+        buffer.putLong(ttl);
+        buffer.putLong(softTtl);
+        DiskBasedCacheUtility.writeHeaderList(allResponseHeaders, buffer);
+    }
+
+    int getHeaderSize() throws IOException {
+        int x = 36;
+        try {
+            x += key.getBytes("UTF8").length;
+            if (etag != null) {
+                x += etag.getBytes("UTF8").length;
+            }
+        } catch (UnsupportedEncodingException e) {
+            VolleyLog.e(e.toString());
+        }
+        x += DiskBasedCacheUtility.headerListSize(allResponseHeaders);
+        return x;
     }
 }
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -1,15 +1,12 @@
 package com.android.volley.toolbox;
 
-import android.os.Build;
 import androidx.annotation.Nullable;
-import androidx.annotation.RequiresApi;
 import com.android.volley.Cache;
 import com.android.volley.Header;
 import com.android.volley.VolleyLog;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 
@@ -152,7 +149,6 @@ class CacheHeader {
     }
 
     /** Writes the contents of this CacheHeader to the specified ByteBuffer. */
-    @RequiresApi(Build.VERSION_CODES.O)
     void writeHeader(ByteBuffer buffer) throws IOException {
         buffer.putInt(CACHE_MAGIC);
         DiskBasedCacheUtility.writeString(buffer, key);
@@ -165,12 +161,11 @@ class CacheHeader {
     }
 
     /** Gets the size of the header in bytes. */
-    @RequiresApi(Build.VERSION_CODES.O)
     int getHeaderSize() throws IOException {
         int size = 0;
-        size += key.getBytes(StandardCharsets.UTF_8).length;
+        size += key.getBytes("UTF-8").length;
         if (etag != null) {
-            size += etag.getBytes(StandardCharsets.UTF_8).length;
+            size += etag.getBytes("UTF-8").length;
         }
         size += DiskBasedCacheUtility.headerListSize(allResponseHeaders);
         return size + HEADER_SIZE;
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -1,13 +1,15 @@
 package com.android.volley.toolbox;
 
+import android.os.Build;
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 import com.android.volley.Cache;
 import com.android.volley.Header;
 import com.android.volley.VolleyLog;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 
@@ -150,6 +152,7 @@ class CacheHeader {
     }
 
     /** Writes the contents of this CacheHeader to the specified ByteBuffer. */
+    @RequiresApi(Build.VERSION_CODES.O)
     void writeHeader(ByteBuffer buffer) throws IOException {
         buffer.putInt(CACHE_MAGIC);
         DiskBasedCacheUtility.writeString(buffer, key);
@@ -162,15 +165,12 @@ class CacheHeader {
     }
 
     /** Gets the size of the header in bytes. */
+    @RequiresApi(Build.VERSION_CODES.O)
     int getHeaderSize() throws IOException {
         int size = 0;
-        try {
-            size += key.getBytes("UTF8").length;
-            if (etag != null) {
-                size += etag.getBytes("UTF8").length;
-            }
-        } catch (UnsupportedEncodingException e) {
-            VolleyLog.e(e, "UTF-8 is unsupported");
+        size += key.getBytes(StandardCharsets.UTF_8).length;
+        if (etag != null) {
+            size += etag.getBytes(StandardCharsets.UTF_8).length;
         }
         size += DiskBasedCacheUtility.headerListSize(allResponseHeaders);
         return size + HEADER_SIZE;
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -146,6 +146,7 @@ class CacheHeader {
         }
     }
 
+    /** Writes the contents of this CacheHeader to the specified ByteBuffer. */
     void writeHeader(ByteBuffer buffer) throws IOException {
         buffer.putInt(CACHE_MAGIC);
         DiskBasedCacheUtility.writeString(buffer, key);
@@ -157,8 +158,9 @@ class CacheHeader {
         DiskBasedCacheUtility.writeHeaderList(allResponseHeaders, buffer);
     }
 
+    /** Gets the size of the header in bytes */
     int getHeaderSize() throws IOException {
-        int x = 36;
+        int x = 44; //4 longs, 3 ints
         try {
             x += key.getBytes("UTF8").length;
             if (etag != null) {
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -161,7 +161,7 @@ class CacheHeader {
         DiskBasedCacheUtility.writeHeaderList(allResponseHeaders, buffer);
     }
 
-    /** Gets the size of the header in bytes */
+    /** Gets the size of the header in bytes. */
     int getHeaderSize() throws IOException {
         int size = 0;
         try {
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -160,7 +160,7 @@ class CacheHeader {
 
     /** Gets the size of the header in bytes */
     int getHeaderSize() throws IOException {
-        int x = 44; //4 longs, 3 ints
+        int x = 44; // 4 longs, 3 ints
         try {
             x += key.getBytes("UTF8").length;
             if (etag != null) {
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -28,7 +28,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
     private final Map<String, CacheHeader> mEntries = new LinkedHashMap<>(16, .75f, true);
 
     /** The supplier for the root directory to use for the cache. */
-    private final DiskBasedCacheUtility.FileSupplierParent mRootDirectorySupplier;
+    private final DiskBasedCacheUtility.FileSupplier mRootDirectorySupplier;
 
     /** Total amount of space currently used by the cache in bytes. */
     private long mTotalSize = 0;
@@ -43,7 +43,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
      */
     public DiskBasedAsyncCache(final File rootDirectory, int maxCacheSizeInBytes) {
         mRootDirectorySupplier =
-                new DiskBasedCacheUtility.FileSupplierParent() {
+                new DiskBasedCacheUtility.FileSupplier() {
                     @Override
                     public File get() {
                         return rootDirectory;
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -28,7 +28,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
     private final Map<String, CacheHeader> mEntries = new LinkedHashMap<>(16, .75f, true);
 
     /** The supplier for the root directory to use for the cache. */
-    private final DiskBasedCacheUtility.FileSupplier mRootDirectorySupplier;
+    private final FileSupplierParent mRootDirectorySupplier;
 
     /** Total amount of space currently used by the cache in bytes. */
     private long mTotalSize = 0;
@@ -43,7 +43,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
      */
     public DiskBasedAsyncCache(final File rootDirectory, int maxCacheSizeInBytes) {
         mRootDirectorySupplier =
-                new DiskBasedCacheUtility.FileSupplier() {
+                new FileSupplierParent() {
                     @Override
                     public File get() {
                         return rootDirectory;
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -61,15 +61,16 @@ public class DiskBasedAsyncCache extends AsyncCache {
             return;
         }
         final File file = DiskBasedCacheUtility.getFileForKey(key, mRootDirectorySupplier);
-        final int size = (int) file.length();
         Path path = Paths.get(file.getPath());
         try {
             final AsynchronousFileChannel afc =
                     AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+            int headerSize = entry.getHeaderSize();
+            final int size = (int) file.length() - headerSize;
             final ByteBuffer buffer = ByteBuffer.allocate(size);
             afc.read(
                     /* destination= */ buffer,
-                    /* position= */ 0,
+                    /* position= */ headerSize,
                     /* attachment= */ null,
                     new CompletionHandler<Integer, Void>() {
                         @Override
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -103,11 +103,12 @@ public class DiskBasedAsyncCache extends AsyncCache {
         // deleted, then skip writing the entry in the first place, as this is just churn.
         // Note that we don't include the cache header overhead in this calculation for simplicity,
         // so putting entries which are just below the threshold may still cause this churn.
-        if (mTotalSize + entry.data.length > mMaxCacheSizeInBytes
-                && entry.data.length
-                        > mMaxCacheSizeInBytes * DiskBasedCacheUtility.HYSTERESIS_FACTOR) {
+        if (DiskBasedCacheUtility.checkPrune(mTotalSize, entry.data.length, mMaxCacheSizeInBytes)
+                && DiskBasedCacheUtility.checkWouldDelete(
+                        entry.data.length, mMaxCacheSizeInBytes)) {
             return;
         }
+
         final File file = DiskBasedCacheUtility.getFileForKey(key, mRootDirectorySupplier);
         Path path = Paths.get(file.getPath());
 
@@ -125,7 +126,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
                     /* attachment= */ null,
                     new CompletionHandler<Integer, Void>() {
                         @Override
-                        public void completed(Integer integer, Void aVoid) {
+                        public void completed(Integer resultLen, Void ignore) {
                             header.size = file.length();
                             mTotalSize =
                                     DiskBasedCacheUtility.putEntry(
@@ -162,11 +163,12 @@ public class DiskBasedAsyncCache extends AsyncCache {
 
     /** Re-initialize the cache if the directory was deleted. */
     private void initializeIfRootDirectoryDeleted() {
-        if (!mRootDirectorySupplier.get().exists()) {
-            VolleyLog.d("Re-initializing cache after external clearing.");
-            mEntries.clear();
-            mTotalSize = 0;
-            initialize();
+        if (mRootDirectorySupplier.get().exists()) {
+            return;
         }
+        VolleyLog.d("Re-initializing cache after external clearing.");
+        mEntries.clear();
+        mTotalSize = 0;
+        initialize();
     }
 }
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -28,7 +28,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
     private final Map<String, CacheHeader> mEntries = new LinkedHashMap<>(16, .75f, true);
 
     /** The supplier for the root directory to use for the cache. */
-    private final DiskBasedCacheUtility.FileSupplier mRootDirectorySupplier;
+    private final DiskBasedCacheUtility.FileSupplierParent mRootDirectorySupplier;
 
     /** Total amount of space currently used by the cache in bytes. */
     private long mTotalSize = 0;
@@ -43,7 +43,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
      */
     public DiskBasedAsyncCache(final File rootDirectory, int maxCacheSizeInBytes) {
         mRootDirectorySupplier =
-                new DiskBasedCacheUtility.FileSupplier() {
+                new DiskBasedCacheUtility.FileSupplierParent() {
                     @Override
                     public File get() {
                         return rootDirectory;
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -212,7 +212,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
             afc.close();
             return true;
         } catch (IOException e) {
-            VolleyLog.e(e, "failed to close file after " + endOfMessage);
+            VolleyLog.e(e, "failed to close file after %s", endOfMessage);
             return false;
         }
     }
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -63,8 +63,9 @@ public class DiskBasedAsyncCache extends AsyncCache {
         final File file = DiskBasedCacheUtility.getFileForKey(key, mRootDirectorySupplier);
         final int size = (int) file.length();
         Path path = Paths.get(file.getPath());
-        try (AsynchronousFileChannel afc =
-                AsynchronousFileChannel.open(path, StandardOpenOption.READ)) {
+        try {
+            final AsynchronousFileChannel afc =
+                    AsynchronousFileChannel.open(path, StandardOpenOption.READ);
             final ByteBuffer buffer = ByteBuffer.allocate(size);
             afc.read(
                     /* destination= */ buffer,
@@ -81,12 +82,22 @@ public class DiskBasedAsyncCache extends AsyncCache {
                                 return;
                             }
                             byte[] data = buffer.array();
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
+                            }
                             callback.onGetComplete(entry.toCacheEntry(data));
                         }
 
                         @Override
                         public void failed(Throwable exc, Void v) {
                             VolleyLog.e(exc, "Failed to read file %s", file.getAbsolutePath());
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
+                            }
                             callback.onGetComplete(null);
                         }
                     });
@@ -122,6 +133,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
             ByteBuffer buffer = ByteBuffer.allocate(size);
             header.writeHeader(buffer);
             buffer.put(entry.data);
+            buffer.flip();
             afc.write(
                     /* source= */ buffer,
                     /* position= */ 0,
@@ -129,7 +141,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
                     new CompletionHandler<Integer, Void>() {
                         @Override
                         public void completed(Integer resultLen, Void ignore) {
-                            header.size = file.length();
+                            header.size = resultLen;
                             mTotalSize =
                                     DiskBasedCacheUtility.putEntry(
                                             key, header, mTotalSize, mEntries);
@@ -139,17 +151,12 @@ public class DiskBasedAsyncCache extends AsyncCache {
                                             mMaxCacheSizeInBytes,
                                             mEntries,
                                             mRootDirectorySupplier);
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
+                            }
                             callback.onPutComplete();
-                            try {
-                                afc.close();
-                            } catch (IOException e) {
-                                VolleyLog.e(e, "Failed to close channel");
-                            }
-                            try {
-                                afc.close();
-                            } catch (IOException e) {
-                                VolleyLog.e(e, "Failed to close channel");
-                            }
                         }
 
                         @Override
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -96,6 +96,7 @@ public class DiskBasedAsyncCache extends AsyncCache {
         }
     }
 
+    /** Puts the cache entry with a specified key into the cache. */
     @Override
     public void put(final String key, Cache.Entry entry, final OnPutCompleteCallback callback) {
         // If adding this entry would trigger a prune, but pruning would cause the new entry to be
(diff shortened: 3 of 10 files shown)


src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +42,85 @@ static String getFilenameForKey(String key) {
         return localFilename;
     }
 
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    /**
+     * Prunes the .
```
sophiahaim, line 51: I think you missed part of the comment here

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +42,85 @@ static String getFilenameForKey(String key) {
...
+     *
+     * @param totalSize The total size of the cache.
+     * @param maxCacheSizeInBytes Maximum size of the cache.
+     * @param entries Map of the entries in the cache.
+     * @param rootDirectorySupplier The supplier for the root directory to use for the cache.
+     * @return A long to update the totalSize.
```
sophiahaim, line 57: nit: This reads a bit strange. Maybe:
@return The updated totalSize

Same for #putEntry below

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +42,85 @@ static String getFilenameForKey(String key) {
...
+        while (iterator.hasNext()) {
+            Map.Entry<String, CacheHeader> entry = iterator.next();
+            CacheHeader e = entry.getValue();
+            boolean deleted = getFileForKey(e.key, rootDirectorySupplier).delete();
+            if (deleted) {
+                totalSize -= e.size;
+            } else {
+                VolleyLog.d(
+                        "Could not delete cache entry for key=%s, filename=%s",
+                        e.key, DiskBasedCacheUtility.getFilenameForKey(e.key));
```
sophiahaim, line 85: Do you need this? You're already in DiskBasedCacheUtility, so I think you can just call #getFilenameForKey

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +42,85 @@ static String getFilenameForKey(String key) {
...
+            }
+            iterator.remove();
+            prunedFiles++;
+
+            if (totalSize < maxCacheSizeInBytes * DiskBasedCacheUtility.HYSTERESIS_FACTOR) {
```
sophiahaim, line 90: Same here – I think you can just use HYSTERESIS_FACTOR directly

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -143,4 +145,31 @@ boolean writeHeader(OutputStream os) {
...
+        buffer.putLong(serverDate);
+        buffer.putLong(lastModified);
+        buffer.putLong(ttl);
+        buffer.putLong(softTtl);
+        DiskBasedCacheUtility.writeHeaderList(allResponseHeaders, buffer);
+    }
+
+    /** Gets the size of the header in bytes */
+    int getHeaderSize() throws IOException {
+        int x = 44; // 4 longs, 3 ints
```
sophiahaim, line 163: make this a CONSTANT

Also, use a clearer variable name instead of x – maybe size?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -104,6 +188,17 @@ static void writeString(OutputStream os, String s) throws IOException {
         os.write(b, 0, b.length);
     }
 
+    static void writeString(ByteBuffer buffer, String s) throws IOException {
```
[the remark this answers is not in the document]
sphill99: added @nullable in all the necessary spots.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +96,77 @@ public void failed(Throwable exc, Void v) {
...
+        }
+    }
+
+    public void initialize() {
+        // TODO (sphill99): #181
+    }
+
+    /** Re-initialize the cache if the directory was deleted. */
+    private void initializeIfRootDirectoryDeleted() {
+        if (!mRootDirectorySupplier.get().exists()) {
```
jbir789, line 165: nit: what do you think of returning early instead? I find that it improves readability as there is less nested code

```java
if (mRootDirectorySupplier.get().exists()) {
  return;
}

// Else do your work here
```

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -143,4 +145,31 @@ boolean writeHeader(OutputStream os) {
...
+        try {
+            x += key.getBytes("UTF8").length;
+            if (etag != null) {
+                x += etag.getBytes("UTF8").length;
+            }
+        } catch (UnsupportedEncodingException e) {
+            VolleyLog.e(e.toString());
```
jbir789, line 170: Could you log the exception with a message instead?

```java
VolleyLog.e(e, "Your message here")
```

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: fixed both of these

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +96,77 @@ public void failed(Throwable exc, Void v) {
...
+    /** Puts the cache entry with a specified key into the cache. */
+    @Override
+    public void put(final String key, Cache.Entry entry, final OnPutCompleteCallback callback) {
+        // If adding this entry would trigger a prune, but pruning would cause the new entry to be
+        // deleted, then skip writing the entry in the first place, as this is just churn.
+        // Note that we don't include the cache header overhead in this calculation for simplicity,
+        // so putting entries which are just below the threshold may still cause this churn.
+        if (mTotalSize + entry.data.length > mMaxCacheSizeInBytes
+                && entry.data.length
+                        > mMaxCacheSizeInBytes * DiskBasedCacheUtility.HYSTERESIS_FACTOR) {
```
jbir789, line 108: Should these `>` signs be replaced with `>=`? Looking at `pruneIfNeeded`, my understanding is that if the new size is bigger or equal to the max size, the cache would be pruned.

jbir789, line 108: It's a good idea to check for this. However, we now have duplicate code in `pruneIfNeeded` and here, in the sense we're checking the same conditions in 2 different places. This is error prone: in the future, if one condition is updated and someone forgets to update the other, it would result in the 2 checks being out of sync and we could end up with not adding entries to the cache when it would have been totally fine.

Instead, I suggest to have your conditions defined in a shared place, most likely in `DiskBasedCacheUtility`. It's OK to split them into 2 methods (one for checking the max cache size, and one with the hysteresis factor), but at least you'd have them defined in one place which will prevent them from being out of sync. There would still be a risk that `pruneIfNeeded` is updated and its logic doesn't correspond to the check here, so it's not perfect, but the chances of getting out of sync are decreased.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +96,77 @@ public void failed(Throwable exc, Void v) {
...
+            ByteBuffer buffer = ByteBuffer.allocate(size);
+            header.writeHeader(buffer);
+            buffer.put(entry.data);
+            afc.write(
+                    /* source= */ buffer,
+                    /* position= */ 0,
+                    /* attachment= */ null,
+                    new CompletionHandler<Integer, Void>() {
+                        @Override
+                        public void completed(Integer integer, Void aVoid) {
```
jbir789, line 128: could you find a better name for `integer` and `aVoid` to be more descriptive? As is, I don't know what they stand for when reading the code.

If they're ignored, you can name them `ignore`, but prefer a descriptive name, especially for the integer.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +42,85 @@ static String getFilenameForKey(String key) {
...
+     * Puts the entry with the specified key into the cache.
+     *
+     * @param key The key to identify the entry by.
+     * @param entry The entry to cache.
+     * @param totalSize The total size of the cache.
+     * @param entries Map of the entries in the cache.
+     * @return A long to update the total size.
+     */
+    static long putEntry(
+            String key, CacheHeader entry, long totalSize, Map<String, CacheHeader> entries) {
```
jbir789, line 113: I understand why you're passing `entries` here, now that it's static, but keep in mind that it's generally not a good practice to modify the arguments of a method, as they're interpreted as an input to the function and not something that will be modified. The user will have to check the doc (which btw should reflect that `entries` will be updated`) to fully understand what the method does (and while it's always a good idea to do so, you can't guarantee future developers will do so, and it's always better to go with self-explicit methods).

I could be convinced that it's fine to keep this as is because it's internal, but we would need a more descriptive documentation. Alternatively, you could keep the method to compute the new total size, and do the `entries.put(key, entry)` at the caller site instead. I would personally prefer this approach.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I'll change these

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Yeah, I agree with that. I just made those changes.

sphill99: I didn't write these checks, so I'm not totally sure why this is, but it does look like the cache is pruned if the total size if the same as the max size. Would be want to change the check, or change in pruneIfNeeded?

src/main/java/com/android/volley/toolbox/CacheHeader.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: Would this be true for pruneIfNeeded as well, since it also changes the entries map? And there will probably be a couple more methods like this (removeEntry, etc.) eventually. Since this would duplicate a lot of code, do you still think that this would be the better route?

Also, we could have the methods return both the entries map and the updated totalSize, then set it in the caller. What're your thoughts?

jbir789: I thought about returning both, but I am not sure it would make the code more readable, so we can stick to this if you think it's better this way.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jbir789: To keep it functionally the same, I'd update these. Note that if you have a method to check those conditions (see the other comment), we wouldn't need to update them because the shared method would take care of keeping the check consistent for us.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
         return localFilename;
     }
 
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    static boolean checkPrune(long totalSize, int dataLength, int maxCacheSize) {
```
jbir789, line 51: nit: the method name isn't super descriptive. What do you think of `wouldExceedCacheSize`?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
...
+                totalSize -= e.size;
+            } else {
+                VolleyLog.d(
+                        "Could not delete cache entry for key=%s, filename=%s",
+                        e.key, getFilenameForKey(e.key));
+            }
+            iterator.remove();
+            prunedFiles++;
+
+            if (totalSize < maxCacheSizeInBytes * HYSTERESIS_FACTOR) {
```
jbir789, line 100: Can you use the new function here?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
...
+     * @param entries Map of the entries in the cache.
+     * @param rootDirectorySupplier The supplier for the root directory to use for the cache.
+     * @return The updated totalSize.
+     */
+    static long pruneIfNeeded(
+            long totalSize,
+            int maxCacheSizeInBytes,
+            Map<String, CacheHeader> entries,
+            FileSupplier rootDirectorySupplier) {
+        if (totalSize < maxCacheSizeInBytes) {
```
jbir789, line 74: Can you use the new function here?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
jbir789, line 51: You could make the function more versatile and really ensure all the checks are in sync by having it take a `newTotalSize` instead of both a `totalSize` and `dataLength`

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
...
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    static boolean checkPrune(long totalSize, int dataLength, int maxCacheSize) {
+        return (totalSize + dataLength > maxCacheSize);
+    }
+
+    static boolean checkWouldDelete(int dataLength, int maxCacheSize) {
```
jbir789, line 55: Could you find a more descriptive name here as well?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: yeah, I like that more

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: I think we can, but should change the methods to <= then. As otherwise it wouldn't cover the same cases.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: changed

sphill99:
I believe I have addressed all comments on this PR

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
...
     }
 
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    static boolean wouldExceedCacheSize(long newTotalSize, int maxCacheSize) {
+        return (newTotalSize >= maxCacheSize);
+    }
```
jbir789, line 53: I think it's safer to keep > rather than >= because that's what would have result in a prune, what do you think? (Same below)

sphill99: yeah, I think that makes sense. I'll change it back.

sphill99: I changed it in the method, but this lead to a test failing. In pruneIfNeeded, it was originally checking if totalSize <= maxCacheSize. To use the method here, I added 1 to totalSize, so to make the check the same as before.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
...
+     * @param entries Map of the entries in the cache.
+     * @param rootDirectorySupplier The supplier for the root directory to use for the cache.
+     * @return The updated totalSize.
+     */
+    static long pruneIfNeeded(
+            long totalSize,
+            int maxCacheSizeInBytes,
+            Map<String, CacheHeader> entries,
+            FileSupplier rootDirectorySupplier) {
+        if (!wouldExceedCacheSize(totalSize + 1, maxCacheSizeInBytes)) {
```
joebowbeer, line 74: This `+1` needs a comment here, at least, and maybe in the method as well. How is the caller suppose to know they should add one?

sphill99: Yeah, I'll add comments explaining this. I did this to essentially make the less than check, a less than or equal to check to match the original code.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
jbir789: Should we update the test instead, by any chance?

joebowbeer: Agree. This code originated from

```
        if (mTotalSize + entry.data.length > mMaxCacheSizeInBytes
                && entry.data.length
                        > mMaxCacheSizeInBytes * DiskBasedCacheUtility.HYSTERESIS_FACTOR) {
            return;
        }
```

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
joebowbeer: The original code here is:

```
if (mTotalSize < mMaxCacheSizeInBytes) {
            return;
        }
```

The comparison here is would be less than, rather than not would exceed. I think the function call isobscuring the logic, and it's not hiding anything data.

The test is correct.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
jbir789: `DiskBasedCache` had the following check in `#put`:

```java
        if (mTotalSize + entry.data.length > mMaxCacheSizeInBytes
                && entry.data.length
                        > mMaxCacheSizeInBytes * DiskBasedCacheUtility.HYSTERESIS_FACTOR) {
          return;
        }
```

sphill99: I don't think it super matters anyway. Since the check in put doesn't count the size of the CacheHeader (aside from the data itself). So if it was already really close to the max, it may get put and then pruned. I switched it to >= since that does make more sense (and then holds for the check in pruneIfNeeded), but I don't think it makes a big difference either way.

jbir789: Sounds good, thanks!

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
         return localFilename;
     }
 
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    static boolean wouldExceedCacheSize(long newTotalSize, int maxCacheSize) {
+        return (newTotalSize >= maxCacheSize);
```
jbir789, line 52: nit: the parens aren't needed

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
...
+                totalSize -= e.size;
+            } else {
+                VolleyLog.d(
+                        "Could not delete cache entry for key=%s, filename=%s",
+                        e.key, getFilenameForKey(e.key));
+            }
+            iterator.remove();
+            prunedFiles++;
+
+            if (!isDataTooLarge((int) totalSize, maxCacheSizeInBytes)) {
```
jbir789, line 100: Casting a `long` to `int` is a lossy conversion: if the value of the long needs more than 32 bits, some of them will be discarded to fit in the `int`.

It'd be safer for `isDataTooLarge` to take a `long` instead (so you'd be casting an `int` to `long`, which is fine)

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,95 @@ static String getFilenameForKey(String key) {
         return localFilename;
     }
 
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    static boolean wouldExceedCacheSize(long newTotalSize, int maxCacheSize) {
```
jbir789, line 51: nit: make `maxCacheSize` a long to be consistent with the total size

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: oh yeah, that makes sense. thanks!

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -24,5 +24,12 @@
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
 
+    public interface OnPutCompleteCallback {
+        /** Invoked when the put to the cache is complete */
```
sophiahaim, line 28: super nit: add a period

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -14,6 +16,9 @@
     /** Magic number for current version of cache file format. */
     private static final int CACHE_MAGIC = 0x20150306;
 
+    /** Bits required to write 4 longs and 3 ints */
```
sophiahaim, line 19: Same here – add a period

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
joebowbeer: The term `exceed` in this method name is misleading as implemented, but I think exceed is correct.

I'll collect comments on these two methods below.

sphill99:
Had to remove the try-with-resources from the get and put methods, as the channel was closing before the reads and writes were actually finished. Same issue with putting the close function in a finally block. Also had to flip the byte buffer after putting the bytes, and needed to start reading at the beginning of the data.

@joebowbeer currently I split it into two methods, so that the pruneIfNeeded method and the put methods don't get out of sync. Since the pruneIfNeeded method does two separate checks, I split it into two different methods. I could make one wouldBePruned method that could make those two checks, and that would clear up exactly what the check at the beginning of put is doing and move the comment to one spot. I'm not sure if that would settle your concerns with the syncing of these methods.

joebowbeer:
> @joebowbeer currently I split it into two methods, so that the pruneIfNeeded method and the put methods don't get out of sync. Since the pruneIfNeeded method does two separate checks, I split it into two different methods. I could make one wouldBePruned method that could make those two checks, and that would clear up exactly what the check at the beginning of put is doing and move the comment to one spot. I'm not sure if that would settle your concerns with the syncing of these methods.

1. The quick-return check in `pruneIfNeeded` should be `totalCacheSize <= maxCacheSize`. It should check if the cache does not violate its size restriction. If not then quit now before pruning. (The missing `=` sign was introduced a couple years ago when the size check was moved from pre-write to post-write.)

2. Aside: I don't think the expression above can be made any clearer. The variables are clearly named. Everyone understands the `<=` operator. Using a method just obfuscates it, esp. if you're passing-in all the arguments, and esp. if the method doesn't even do what it proclaims to do. It'sessentially a glorified `greaterThanOrEqual` method that is misnamed `greaterThan`.

3. Yes, I think `wouldBePruned` is a good name for the method that can be shared between its two users. Please unit test this, particularly around the int/long boundary and around the maxCacheSize boundary.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -76,12 +83,22 @@ public void completed(Integer result, Void v) {
                                 return;
                             }
                             byte[] data = buffer.array();
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
```
jbir789, line 89: super small nit: having more descriptive error logs can be helpful to narrow down where the error comes from. This is why I'd suggest to say something like `Failed to close channel after successful read` here, and `Failed to close channel after failed read attempt` below?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +108,92 @@ public void failed(Throwable exc, Void v) {
...
+                            mTotalSize =
+                                    DiskBasedCacheUtility.pruneIfNeeded(
+                                            mTotalSize,
+                                            mMaxCacheSizeInBytes,
+                                            mEntries,
+                                            mRootDirectorySupplier);
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
```
jbir789, line 158: Same for this log

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +108,92 @@ public void failed(Throwable exc, Void v) {
...
+
+                        @Override
+                        public void failed(Throwable throwable, Void aVoid) {
+                            VolleyLog.e(
+                                    throwable, "Failed to read file %s", file.getAbsolutePath());
+                            callback.onPutComplete();
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
```
jbir789, line 171: And this one

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +108,92 @@ public void failed(Throwable exc, Void v) {
...
+                            }
+                            callback.onPutComplete();
+                        }
+
+                        @Override
+                        public void failed(Throwable throwable, Void aVoid) {
```
jbir789, line 164: nit: can you rename `aVoid` to `ignore`?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -76,12 +83,22 @@ public void completed(Integer result, Void v) {
...
                         public void failed(Throwable exc, Void v) {
                             VolleyLog.e(exc, "Failed to read file %s", file.getAbsolutePath());
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
+                            }
                             callback.onGetComplete(null);
                         }
                     });
```
jbir789, line 104: Should we close the afc in the catch below as well? If we failed to read the file, and respond to the callback, I don't expect we still need the channel to be open (assuming we were able to open it)

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -91,8 +108,92 @@ public void failed(Throwable exc, Void v) {
...
+                            callback.onPutComplete();
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel");
+                            }
+                        }
+                    });
+        } catch (IOException e) {
+            boolean deleted = file.delete();
```
jbir789, line 176: Same here, should we close the channel if it was open?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: It doesn't let me add it there, since the channel is defined in the try block. And I can't declare the channel outside of it, since it needs to be final. Is there any way around this? I think we should be closing it, since IOExceptions can be thrown between opening the channel and performing the read/write.

joebowbeer:
@jbir789 My primary request is to consolidate the `wouldBePruned` method, for the comment if for no other reason.

As for the test in `pruneIfNeeded`, it is potentially a different test, since it is performed **after** the cache is written, so costs are different. Also, consolidating `wouldBePruned` may reduce the number of call sites below three (and so rule-of-three no longer applies?). More to the point, I think that creating a method for `>` is a false economy. Regardless of the shared method and what its name implies, the right arguments need to be supplied in the right order. Adding a method makes the code more difficult to review at the local level, e.g., if a change is introduced in only one location where the method body is not visible. I don't think it helps, but I can't object if the method implementation matches its name and if the code is effectively tested.

I do think the quick-return test in `pruneIfNeeded` should `<=`. That does not need to be part of this PR, but if the code is being changed anyway...

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
joebowbeer: Strictly speaking the final catch(IOE) should close the channel if it is open. To better isolate the different failures, I would add a try/finally after opening the channel. Where the finally block closes the channel. Then outer try/catch can just report the error. (See pattern in DiskBasedCache.)

OTOH, the afc.read can throw IOE (NonReadableChannelException) only if channel is not open for reading, so you might never need to close in this exceptional case, but comments are needed to explain, and may also be needed to disable static checks.

sphill99: I tried it with a finally block, but with an AsynchronousFileChannel, it will get closed before finishing the read/write, and then cause an error. And there can be an IOException thrown while writing the CacheHeader (UnsupportedEncodingException), so that would be a case where the channel wouldn't actually be closed. I think closing the channel in the catch block may be the only way to do it.

joebowbeer: @sphill99  I see what you mean now

sphill99:
I believe I've now addressed all comments on this PR. Let me know if I missed any, or didn't make the necessary changes.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -76,23 +87,129 @@ public void completed(Integer result, Void v) {
...
+            buffer.flip();
+            afc.write(
+                    /* source= */ buffer,
+                    /* position= */ 0,
+                    /* attachment= */ null,
+                    new CompletionHandler<Integer, Void>() {
+                        @Override
+                        public void completed(Integer resultLen, Void ignore) {
+                            header.size = resultLen;
+                            mTotalSize =
```
jpd236, line 155: It looks a little peculiar to have:

mTotalSize = x;
mTotalSize = y;

rather than just
x;
mTotalSize = y;

since the first set is immediately clobbered by the second.

That seems worth fixing, but it does actually raise a question that I don't think we've considered yet. What thread safety guarantees should this cache provide? The old cache had "synchronized" keywords over all public API methods, and since the methods are blocking, that ensures thread safety by ensuring only one operation can occur at a time on the entire cache. The new implementation doesn't appear to do anything to address thread safety, and synchronized keywords wouldn't be enough since the disk operations are now asynchronous. I also think we need to be careful about when we try to grab the lock - if we try to grab it before triggering an async operation, and the lock will be blocked on some other async operation, then we effectively make the non-blocking API blocking. So we'll need to think through solutions here.

I wouldn't block this PR on it - just something we'll have to address in a follow-up.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -76,23 +87,129 @@ public void completed(Integer result, Void v) {
...
+                                    DiskBasedCacheUtility.pruneIfNeeded(
+                                            mTotalSize,
+                                            mMaxCacheSizeInBytes,
+                                            mEntries,
+                                            mRootDirectorySupplier);
+                            try {
+                                afc.close();
+                            } catch (IOException e) {
+                                VolleyLog.e(e, "Failed to close channel after successful write");
```
jpd236, line 167: I don't think the write can be considered successful if the close fails - e.g. I don't think we should update total size or perform pruning until the close completes, and then only if it's successful.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jpd236: It seems like we should try to close it before trying to delete the file. I would imagine that deleting a file and then calling close() could cause the file to reappear, but not sure.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -241,8 +236,10 @@ public synchronized void put(String key, Entry entry) {
             fos.write(entry.data);
             fos.close();
             e.size = file.length();
-            putEntry(key, e);
-            pruneIfNeeded();
+            mTotalSize = DiskBasedCacheUtility.putEntry(key, e, mTotalSize, mEntries);
```
jpd236, line 239: remove "mTotalSize = " - no need to set this twice, and the parent method and I believe all other reads of mTotalSize are synchronized so mTotalSize shouldn't be observable between these two sets.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +43,120 @@ static String getFilenameForKey(String key) {
         return localFilename;
     }
 
+    /** Returns a file object for the given cache key. */
+    public static File getFileForKey(String key, FileSupplier rootDirectorySupplier) {
+        return new File(rootDirectorySupplier.get(), getFilenameForKey(key));
+    }
+
+    static boolean wouldExceedCacheSize(long newTotalSize, long maxCacheSize) {
+        return (newTotalSize > maxCacheSize);
```
jpd236, line 52: nit: these parentheses are unnecessary

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -136,4 +281,18 @@ static void writeHeaderList(List<Header> headers, OutputStream os) throws IOExce
...
     }
+
+    static int headerListSize(@Nullable List<Header> headers) throws IOException {
+        if (headers == null) {
+            return 4;
+        }
+        int bytes = 4;
+
+        for (Header header : headers) {
+            bytes += header.getName().getBytes("UTF-8").length;
```
jpd236, line 292: we use UTF-8 here and UTF8 elsewhere. Can we be consistent? Why do we check for UnsupportedEncodingException there and not here?

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
joebowbeer: Disagree. The updated value of mTotalSize is passed to pruneIfNeeded. (I hope a unit test would catch this...)

By the way, having to pass all of these values to the tool method, in the right order, and then update the instance fields, seems brittle. Can't OOP offer more assistance more here?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -76,23 +87,129 @@ public void completed(Integer result, Void v) {
...
+            buffer.put(entry.data);
+            buffer.flip();
+            afc.write(
+                    /* source= */ buffer,
+                    /* position= */ 0,
+                    /* attachment= */ null,
+                    new CompletionHandler<Integer, Void>() {
+                        @Override
+                        public void completed(Integer resultLen, Void ignore) {
+                            header.size = resultLen;
```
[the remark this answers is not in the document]
sphill99: yeah, I think that'd be a good check to make

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
joebowbeer: The first value is used to compute the second.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
sphill99: but it is passed into the next method, so doesn't it need to be updated there?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: Would it make sense to require newer api for these methods, then use StandardCharsets.UTF-8? It's a nio class, and never will have an UEE.

joebowbeer: UTF-8 is the default and is always available

sphill99: I was getting compilation errors unless I specified UTF-8

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
sphill99: Yeah, I was thinking about better ways to do this. I think it's good to share the code, since it'd be a lot of code to be duplicated. I don't think this is terrible, since the methods take in all different types, which means it'll tell you if the arguments aren't in the right order. The only other method I could think of is passing in the Cache itself. And then making those variables accessible. But I think that may cause more issues.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: if the close fails, would we still try to delete the file?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -76,23 +87,129 @@ public void completed(Integer result, Void v) {
                                 return;
                             }
                             byte[] data = buffer.array();
+                            try {
```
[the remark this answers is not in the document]
sphill99: I added a helper method to make this cleaner. I wasn't sure if this was exactly what you were intending though. Let me know if I should change anything about this.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
jpd236: I don't think we can use StandardCharsets here because this code is shared with the DiskBasedCache code path, which needs more backwards compatibility. Just want to make sure we're consistent with the naming.

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
jpd236: Yeah, sorry, didn't realize it was passed from the first method to the second. Never mind here.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: Okay, I just changed this back. They should all now be "UTF-8"

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -24,5 +24,12 @@
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
 
+    public interface OnPutCompleteCallback {
+        /** Invoked when the put to the cache is complete. */
+        void onPutComplete();
+    }
+
+    public abstract void put(String key, Cache.Entry entry, OnPutCompleteCallback callback);
```
jpd236, line 32: could you add Javadoc for this (like get)

src/main/java/com/android/volley/toolbox/CacheHeader.java
jpd236: fyi, looks like this is still unaddressed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -80,19 +92,119 @@ public void completed(Integer result, Void v) {
...
+                    /* source= */ buffer,
+                    /* position= */ 0,
+                    /* attachment= */ null,
+                    new CompletionHandler<Integer, Void>() {
+                        @Override
+                        public void completed(Integer resultLen, Void ignore) {
+                            if (closeChannel(afc, "completed read")) {
+                                if (resultLen != size) {
+                                    VolleyLog.e(
+                                            "File changed while reading: %s",
```
jpd236, line 143: this is a write, correct?

Also, I think we should delete the file in this case, since it wouldn't have valid contents.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -80,19 +92,119 @@ public void completed(Integer result, Void v) {
...
+            buffer.put(entry.data);
+            buffer.flip();
+            afc.write(
+                    /* source= */ buffer,
+                    /* position= */ 0,
+                    /* attachment= */ null,
+                    new CompletionHandler<Integer, Void>() {
+                        @Override
+                        public void completed(Integer resultLen, Void ignore) {
+                            if (closeChannel(afc, "completed read")) {
```
jpd236, line 140: write

jpd236, line 140: and I think we should delete after a failed close as well, yes. in the existing implementation we delete in the event of any IO failure.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -80,19 +92,119 @@ public void completed(Integer result, Void v) {
...
+        if (mRootDirectorySupplier.get().exists()) {
+            return;
+        }
+        VolleyLog.d("Re-initializing cache after external clearing.");
+        mEntries.clear();
+        mTotalSize = 0;
+        initialize();
+    }
+
+    private boolean closeChannel(AsynchronousFileChannel afc, String endOfMessage) {
```
jpd236, line 198: `@Nullable` afc

jpd236, line 198: consider a quick Javadoc, since the meaning of the return value is perhaps non-obvious.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: fixed

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: added javadoc

src/main/java/com/android/volley/AsyncCache.java
sphill99: added javadoc here

sphill99:
I believe I've addressed all comments currently on this PR.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -80,19 +93,136 @@ public void completed(Integer result, Void v) {
...
+     */
+    private boolean closeChannel(@Nullable AsynchronousFileChannel afc, String endOfMessage) {
+        if (afc == null) {
+            return false;
+        }
+        try {
+            afc.close();
+            return true;
+        } catch (IOException e) {
+            VolleyLog.e(e, "failed to close file after " + endOfMessage);
```
jbir789, line 215: nit: use the String format type of log: `log(e. "failed to close file after %s", endOfMessage)`

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -83,7 +81,8 @@ public File get() {
      *     briefly exceed this size on disk when writing a new entry that pushes it over the limit
      *     until the ensuing pruning completes.
      */
-    public DiskBasedCache(FileSupplier rootDirectorySupplier, int maxCacheSizeInBytes) {
+    public DiskBasedCache(
+            DiskBasedCacheUtility.FileSupplier rootDirectorySupplier, int maxCacheSizeInBytes) {
```
jpd236, line 85: Actually, this is an unsafe API change. This is a public constructor (so Volley clients may have used this), so we can't change the signature. (Moreover, DiskBasedCacheUtility is package private, so external callers can't even access it).

Unfortunately I think we need to leave FileSupplier here and duplicate it for the Async layer.

sphill99: This should be fixed now that FileSupplier is duplicated in DiskBasedCache

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -255,31 +252,28 @@ public synchronized void put(String key, Entry entry) {
...
             VolleyLog.d(
                     "Could not delete cache entry for key=%s, filename=%s",
-                    key, getFilenameForKey(key));
+                    key, DiskBasedCacheUtility.getFilenameForKey(key));
         }
     }
 
+    /** Represents a supplier for {@link File}s. */
+    public interface FileSupplier extends DiskBasedCacheUtility.FileSupplier {
+        @Override
```
jpd236, line 266: Is this necessary? I think you can just do "extends DiskBasedCacheUtility.FileSupplier { }" since you're not changing anything about the underlying method.

sphill99: Oh, I'll move the File Supplier. And you're right,  that isn't necessary,  I'll change that as well

sphill99: I had to change the name to FileSupplierParent, as otherwise, it couldn't be extended by FileSupplier

jpd236: Can you call it FileSupplier and do "public interface FileSupplier implements com.google.android.volley.toolbox.FileSupplier"?

sphill99: Yep, just tried that and it works!

PR #349, opened 2020-07-06
Implementations for remaining async cache methods (initialize, invalidate, remove)

sphill99:
Implemented the remaining methods for this class. This involved adding read methods that took in an AsyncFileChannel. I used a Future for these operations in place of a CompletionHandler, as that seemed much simpler in this case. Consolidated the callback interface for put, init, invalidate, and remove, since they all did the same thing. Can change this back to them each having their own if there are any benefits to that.

Initialize still has a loop over a files array, but couldn't figure out how to do that asynchronously since our map is not thread safe.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedCache.java
@@ -222,9 +222,9 @@ public class DiskBasedCache implements Cache {
         // deleted, then skip writing the entry in the first place, as this is just churn.
         // Note that we don't include the cache header overhead in this calculation for simplicity,
         // so putting entries which are just below the threshold may still cause this churn.
-        if (mTotalSize + entry.data.length > mMaxCacheSizeInBytes
-                && entry.data.length
-                        > mMaxCacheSizeInBytes * DiskBasedCacheUtility.HYSTERESIS_FACTOR) {
+        if (DiskBasedCacheUtility.checkPrune(mTotalSize, entry.data.length, mMaxCacheSizeInBytes)
+                && DiskBasedCacheUtility.checkWouldDelete(
+                        entry.data.length, mMaxCacheSizeInBytes)) {
             return;
         }
         File file = DiskBasedCacheUtility.getFileForKey(key, mRootDirectorySupplier);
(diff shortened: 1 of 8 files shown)


src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -37,6 +48,104 @@ static String getFilenameForKey(String key) {
...
+            totalSize += entry.size;
+        } else {
+            CacheHeader oldEntry = entries.get(key);
+            totalSize += (entry.size - oldEntry.size);
+        }
+        entries.put(key, entry);
+        return totalSize;
+    }
+
+    /** Removes the entry identified by 'key' from the cache. This modifies the entries map */
```
sophiahaim, line 140: Add @return The updated totalSize

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
```diff
@@ -255,31 +256,18 @@ public synchronized void put(String key, Entry entry) {
     /** Removes the specified key from the cache if it exists. */
     @Override
     public synchronized void remove(String key) {
-        boolean deleted = getFileForKey(key).delete();
+        boolean deleted = DiskBasedCacheUtility.getFileForKey(key, mRootDirectorySupplier).delete();
         removeEntry(key);
```
sophiahaim, line 260: Can you use the new DiskBasedCacheUtility#removeEntry function here?

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -136,4 +329,39 @@ static void writeHeaderList(List<Header> headers, OutputStream os) throws IOExce
...
+    @RequiresApi(api = Build.VERSION_CODES.O)
+    static List<Header> readHeaderList(AsynchronousFileChannel channel, int offset)
+            throws IOException {
+        int size = readInt(channel, offset);
+        if (size < 0) {
+            throw new IOException("readHeaderList size=" + size);
+        }
+        offset += 4;
+        List<Header> result =
+                (size == 0) ? Collections.<Header>emptyList() : new ArrayList<Header>();
```
sophiahaim, line 342: nit: don't need the parentheses

src/main/java/com/android/volley/toolbox/DiskBasedCache.java
sphill99: Oh yeah. It looks like I didn't make those changes for DiskBasedCache on this branch.

joebowbeer: We should hold off reviewing this PR until the other PR on which it is based is merged?

sphill99: Yeah, I think that's smart

sphill99:
I didn't realize how much overlap there was between this PR and the PutImpl one. So I'm making this a draft for now.

PR #350, opened 2020-07-17
Implementations for rest of AsyncCache, tests for cache methods, still no thread safety yet

sphill99:
I've implemented the remaining methods in AsyncCache (clear, initialize, remove, invalidate), added a readHeader method to CacheHeader (for validity check in get, and use in initialize), and added tests that test same things as in DiskBasedCacheTest.

Few things to point out about this PR:
- I made a generalized callback interface, since it was the exact same for put, clear, initialize, remove and invalidate. At this point, I didn't see a need for each one to have their own interface.
- Invalidate throws ExecutionException & InterruptedException, since I use completable futures there. Currently I just log the error, but wasn't sure if something more needed to be done in that instance. Also could not use futures, and put code in callbacks, but that seemed a lot less readable.
- I changed helper method closeChannel to return true in the case the channel was null, since no error was thrown.
- I did not include thread safety fixes on this PR. I am still working on finishing up that code, and was going to submit them in a separate PR.

### DIFF
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -39,14 +39,21 @@ public abstract class AsyncCache {
      */
     public abstract void put(String key, Cache.Entry entry, OnCompleteCallback callback);
 
-    /** Clears the cache. Deletes all cached files from disk. */
+    /**
+     * Clears the cache. Deletes all cached files from disk. Calls {@link
+     * OnCompleteCallback#onComplete} after the operation is finished.
+     */
     public abstract void clear(OnCompleteCallback callback);
 
-    /** Initializes the cache. */
+    /**
+     * Initializes the cache and calls {@link OnCompleteCallback#onComplete} after the operation is
+     * finished.
+     */
     public abstract void initialize(OnCompleteCallback callback);
 
     /**
-     * Invalidates an entry in the cache.
+     * Invalidates an entry in the cache and calls {@link OnCompleteCallback#onComplete} after the
+     * operation is finished.
      *
      * @param key Cache key
      * @param fullExpire True to fully expire the entry, false to soft expire
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -24,49 +24,50 @@ public abstract class AsyncCache {
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
 
-    public interface OnCompleteCallback {
+    public interface OnWriteCompleteCallback {
         /** Invoked when the cache operation is complete */
-        void onComplete();
+        void onWriteComplete();
     }
 
     /**
-     * Writes a {@link Cache.Entry} to the cache, and calls {@link OnCompleteCallback#onComplete}
-     * after the operation is finished.
+     * Writes a {@link Cache.Entry} to the cache, and calls {@link
+     * OnWriteCompleteCallback#onWriteComplete} after the operation is finished.
      *
      * @param key Cache key
      * @param entry The entry to be written to the cache
      * @param callback Callback that will be notified when the information has been written
      */
-    public abstract void put(String key, Cache.Entry entry, OnCompleteCallback callback);
+    public abstract void put(String key, Cache.Entry entry, OnWriteCompleteCallback callback);
 
     /**
      * Clears the cache. Deletes all cached files from disk. Calls {@link
-     * OnCompleteCallback#onComplete} after the operation is finished.
+     * OnWriteCompleteCallback#onWriteComplete} after the operation is finished.
      */
-    public abstract void clear(OnCompleteCallback callback);
+    public abstract void clear(OnWriteCompleteCallback callback);
 
     /**
-     * Initializes the cache and calls {@link OnCompleteCallback#onComplete} after the operation is
-     * finished.
-     */
-    public abstract void initialize(OnCompleteCallback callback);
-
-    /**
-     * Invalidates an entry in the cache and calls {@link OnCompleteCallback#onComplete} after the
+     * Initializes the cache and calls {@link OnWriteCompleteCallback#onWriteComplete} after the
      * operation is finished.
+     */
+    public abstract void initialize(OnWriteCompleteCallback callback);
+
+    /**
+     * Invalidates an entry in the cache and calls {@link OnWriteCompleteCallback#onWriteComplete}
+     * after the operation is finished.
      *
      * @param key Cache key
      * @param fullExpire True to fully expire the entry, false to soft expire
      * @param callback Callback that's invoked once the entry has been invalidated
      */
-    public abstract void invalidate(String key, boolean fullExpire, OnCompleteCallback callback);
+    public abstract void invalidate(
+            String key, boolean fullExpire, OnWriteCompleteCallback callback);
 
     /**
-     * Removes a {@link Cache.Entry} from the cache, and calls {@link OnCompleteCallback#onComplete}
-     * after the operation is finished.
+     * Removes a {@link Cache.Entry} from the cache, and calls {@link
+     * OnWriteCompleteCallback#onWriteComplete} after the operation is finished.
      *
      * @param key Cache key
      * @param callback Callback that's invoked once the entry has been removed
      */
-    public abstract void remove(String key, OnCompleteCallback callback);
+    public abstract void remove(String key, OnWriteCompleteCallback callback);
 }
--- a/src/main/java/com/android/volley/AsyncCache.java
+++ b/src/main/java/com/android/volley/AsyncCache.java
@@ -24,20 +24,42 @@ public abstract class AsyncCache {
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
 
-    public interface OnPutCompleteCallback {
-        /** Invoked when the put to the cache is complete. */
-        void onPutComplete();
+    public interface OnCompleteCallback {
+        /** Invoked when the cache operation is complete */
+        void onComplete();
     }
 
     /**
-     * Writes a {@link Cache.Entry} to the cache, and calls {@link
-     * OnPutCompleteCallback#onPutComplete} after the operation is finished.
+     * Writes a {@link Cache.Entry} to the cache, and calls {@link OnCompleteCallback#onComplete}
+     * after the operation is finished.
      *
      * @param key Cache key
      * @param entry The entry to be written to the cache
      * @param callback Callback that will be notified when the information has been written
      */
-    public abstract void put(String key, Cache.Entry entry, OnPutCompleteCallback callback);
+    public abstract void put(String key, Cache.Entry entry, OnCompleteCallback callback);
 
-    // TODO(#181): Implement the rest.
+    /** Clears the cache. Deletes all cached files from disk. */
+    public abstract void clear(OnCompleteCallback callback);
+
+    /** Initializes the cache. */
+    public abstract void initialize(OnCompleteCallback callback);
+
+    /**
+     * Invalidates an entry in the cache.
+     *
+     * @param key Cache key
+     * @param fullExpire True to fully expire the entry, false to soft expire
+     * @param callback Callback that's invoked once the entry has been invalidated
+     */
+    public abstract void invalidate(String key, boolean fullExpire, OnCompleteCallback callback);
+
+    /**
+     * Removes a {@link Cache.Entry} from the cache, and calls {@link OnCompleteCallback#onComplete}
+     * after the operation is finished.
+     *
+     * @param key Cache key
+     * @param callback Callback that's invoked once the entry has been removed
+     */
+    public abstract void remove(String key, OnCompleteCallback callback);
 }
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -1,19 +1,14 @@
 package com.android.volley.toolbox;
 
-import android.os.Build;
 import androidx.annotation.Nullable;
-import androidx.annotation.RequiresApi;
 import com.android.volley.Cache;
 import com.android.volley.Header;
 import com.android.volley.VolleyLog;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
-import java.nio.channels.AsynchronousFileChannel;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
 
 /** Handles holding onto the cache headers for an entry. */
 class CacheHeader {
@@ -116,43 +111,6 @@ class CacheHeader {
         long ttl = DiskBasedCacheUtility.readLong(is);
         long softTtl = DiskBasedCacheUtility.readLong(is);
         List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(is);
-        return new CacheHeader(
-                key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
-    }
-
-    /**
-     * Reads the header from a AsynchronousFileChannel and returns a CacheHeader object.
-     *
-     * @param afc File channel to read from.
-     * @param size Size of the cache header.
-     * @throws IOException if fails to read header
-     */
-    @RequiresApi(api = Build.VERSION_CODES.O)
-    static CacheHeader readHeader(final AsynchronousFileChannel afc, int size) throws IOException {
-        final ByteBuffer buffer = ByteBuffer.allocate(size);
-        Future<Integer> fillBuffer = afc.read(buffer, 0);
-        try {
-            fillBuffer.get();
-        } catch (ExecutionException e) {
-            VolleyLog.e(e.toString(), "ExecutionException thrown");
-            throw new IOException(e);
-        } catch (InterruptedException e) {
-            VolleyLog.e(e.toString(), "InterruptedException caught");
-            Thread.currentThread().interrupt();
-        }
-        buffer.flip();
-        int magic = buffer.getInt();
-        if (magic != CACHE_MAGIC) {
-            // don't bother deleting, it'll get pruned eventually
-            throw new IOException();
-        }
-        String key = DiskBasedCacheUtility.readString(buffer);
-        String etag = DiskBasedCacheUtility.readString(buffer);
-        long serverDate = buffer.getLong();
-        long lastModified = buffer.getLong();
-        long ttl = buffer.getLong();
-        long softTtl = buffer.getLong();
-        List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(buffer);
         return new CacheHeader(
                 key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
     }
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -157,6 +157,36 @@ class CacheHeader {
                 key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
     }
 
+    /**
+     * Reads the header from a AsynchronousFileChannel and returns a CacheHeader object.
+     *
+     * @param buffer Buffer to get header info from.
+     * @throws IOException if fails to read header
+     */
+    @Nullable
+    static CacheHeader readHeader(final ByteBuffer buffer) {
+        try {
+            buffer.flip();
+            int magic = buffer.getInt();
+            if (magic != CACHE_MAGIC) {
+                // don't bother deleting, it'll get pruned eventually
+                return null;
+            }
+            String key = DiskBasedCacheUtility.readString(buffer);
+            String etag = DiskBasedCacheUtility.readString(buffer);
+            long serverDate = buffer.getLong();
+            long lastModified = buffer.getLong();
+            long ttl = buffer.getLong();
+            long softTtl = buffer.getLong();
+            List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(buffer);
+            return new CacheHeader(
+                    key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
+        } catch (IOException e) {
+            VolleyLog.d(e.toString(), "Failed to read CacheHeader");
+            return null;
+        }
+    }
+
     /** Creates a cache entry for the specified data. */
     Cache.Entry toCacheEntry(byte[] data) {
         Cache.Entry e = new Cache.Entry();
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -128,6 +128,7 @@ class CacheHeader {
      * @throws IOException if fails to read header
      */
     @RequiresApi(api = Build.VERSION_CODES.O)
+    @Nullable
     static CacheHeader readHeader(final AsynchronousFileChannel afc, int size) throws IOException {
         final ByteBuffer buffer = ByteBuffer.allocate(size);
         Future<Integer> fillBuffer = afc.read(buffer, 0);
@@ -149,7 +150,7 @@ class CacheHeader {
             return new CacheHeader(
                     key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
         } catch (ExecutionException | InterruptedException e) {
-            throw new IOException(e);
+            return null;
         }
     }
 
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -137,9 +137,8 @@ class CacheHeader {
             VolleyLog.e(e.toString(), "ExecutionException thrown");
             throw new IOException(e);
         } catch (InterruptedException e) {
-            VolleyLog.e(e.toString(), "InterruptedException thrown");
+            VolleyLog.e(e.toString(), "InterruptedException caught");
             Thread.currentThread().interrupt();
-            throw new IOException(e);
         }
         buffer.flip();
         int magic = buffer.getInt();
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -6,6 +6,7 @@ import com.android.volley.Header;
 import com.android.volley.VolleyLog;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.List;
@@ -116,7 +117,7 @@ class CacheHeader {
     }
 
     /**
-     * Reads the header from a AsynchronousFileChannel and returns a CacheHeader object.
+     * Reads the header from a ByteBuffer and returns a CacheHeader object.
      *
      * @param buffer Buffer to get header info from.
      * @throws IOException if fails to read header
@@ -124,10 +125,8 @@ class CacheHeader {
     @Nullable
     static CacheHeader readHeader(final ByteBuffer buffer) {
         try {
-            buffer.flip();
             int magic = buffer.getInt();
             if (magic != CACHE_MAGIC) {
-                // don't bother deleting, it'll get pruned eventually
                 return null;
             }
             String key = DiskBasedCacheUtility.readString(buffer);
@@ -140,7 +139,10 @@ class CacheHeader {
             return new CacheHeader(
                     key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
         } catch (IOException e) {
-            VolleyLog.d(e.toString(), "Failed to read CacheHeader");
+            VolleyLog.e(e, "Failed to read CacheHeader");
+            return null;
+        } catch (BufferUnderflowException e) {
+            VolleyLog.e(e, "Ran out of room while reading from the buffer");
             return null;
         }
     }
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -1,14 +1,19 @@
 package com.android.volley.toolbox;
 
+import android.os.Build;
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 import com.android.volley.Cache;
 import com.android.volley.Header;
 import com.android.volley.VolleyLog;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.nio.channels.AsynchronousFileChannel;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
 /** Handles holding onto the cache headers for an entry. */
 class CacheHeader {
@@ -113,6 +118,41 @@ class CacheHeader {
         List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(is);
         return new CacheHeader(
                 key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
+    }
+
+    /**
+     * Reads the header from a AsynchronousFileChannel and returns a CacheHeader object.
+     *
+     * @param afc File channel to read from.
+     * @param size Size of the cache header.
+     * @throws IOException if fails to read header
+     */
+    @RequiresApi(api = Build.VERSION_CODES.O)
+    static CacheHeader readHeader(final AsynchronousFileChannel afc, int size) throws IOException {
+        final ByteBuffer buffer = ByteBuffer.allocate(size);
+        Future<Integer> fillBuffer = afc.read(buffer, 0);
+        try {
+            fillBuffer.get();
+            buffer.flip();
+            int magic = buffer.getInt();
+            if (magic != CACHE_MAGIC) {
+                // don't bother deleting, it'll get pruned eventually
+                throw new IOException();
+            }
+            String key = DiskBasedCacheUtility.readString(buffer);
+            String etag = DiskBasedCacheUtility.readString(buffer);
+            long serverDate = buffer.getLong();
+            long lastModified = buffer.getLong();
+            long ttl = buffer.getLong();
+            long softTtl = buffer.getLong();
+            List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(buffer);
+            return new CacheHeader(
+                    key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
+        } catch (ExecutionException e) {
+            throw new IOException();
+        } catch (InterruptedException e) {
+            throw new IOException();
+        }
     }
 
     /** Creates a cache entry for the specified data. */
--- a/src/main/java/com/android/volley/toolbox/CacheHeader.java
+++ b/src/main/java/com/android/volley/toolbox/CacheHeader.java
@@ -128,30 +128,34 @@ class CacheHeader {
      * @throws IOException if fails to read header
      */
     @RequiresApi(api = Build.VERSION_CODES.O)
-    @Nullable
     static CacheHeader readHeader(final AsynchronousFileChannel afc, int size) throws IOException {
         final ByteBuffer buffer = ByteBuffer.allocate(size);
         Future<Integer> fillBuffer = afc.read(buffer, 0);
         try {
             fillBuffer.get();
-            buffer.flip();
-            int magic = buffer.getInt();
-            if (magic != CACHE_MAGIC) {
-                // don't bother deleting, it'll get pruned eventually
-                throw new IOException();
-            }
-            String key = DiskBasedCacheUtility.readString(buffer);
-            String etag = DiskBasedCacheUtility.readString(buffer);
-            long serverDate = buffer.getLong();
-            long lastModified = buffer.getLong();
-            long ttl = buffer.getLong();
-            long softTtl = buffer.getLong();
-            List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(buffer);
-            return new CacheHeader(
-                    key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
-        } catch (ExecutionException | InterruptedException e) {
-            return null;
+        } catch (ExecutionException e) {
+            VolleyLog.e(e.toString(), "ExecutionException thrown");
+            throw new IOException(e);
+        } catch (InterruptedException e) {
+            VolleyLog.e(e.toString(), "InterruptedException thrown");
+            Thread.currentThread().interrupt();
+            throw new IOException(e);
         }
+        buffer.flip();
+        int magic = buffer.getInt();
+        if (magic != CACHE_MAGIC) {
+            // don't bother deleting, it'll get pruned eventually
+            throw new IOException();
+        }
+        String key = DiskBasedCacheUtility.readString(buffer);
+        String etag = DiskBasedCacheUtility.readString(buffer);
+        long serverDate = buffer.getLong();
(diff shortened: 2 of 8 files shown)


src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,449 @@
...
+        Cache.Entry entry = randomData(MAX_SIZE - getEntrySizeOnDisk("maxsize") - 1);
+        putEntry("maxsize", entry).get();
+
+        assertThatEntriesAreEqual(getEntry("maxsize").get(), entry);
+
+        // Now any new entry should cause the first one to be cleared.
+        entry = randomData(0);
+        putEntry("bit", entry).get();
+
+        assertThat(getEntry("goodsize").get(), is(nullValue()));
```
sophiahaim, line 122: I'm a little confused by this one. Shouldn't it be checking the "maxsize" entry?

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,449 @@
...
+
+    @Test
+    public void testPutClearGet() throws ExecutionException, InterruptedException {
+        Cache.Entry entry = randomData(511);
+        putEntry("key", entry).get();
+
+        assertThatEntriesAreEqual(getEntry("key").get(), entry);
+
+        cache.clear(
+                new AsyncCache.OnCompleteCallback() {
```
sophiahaim, line 272: Why did you decide to do the asserts in the callback this time instead of using a future?

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    @Test
+    public void testInvalidate() throws ExecutionException, InterruptedException {
+        Cache.Entry entry = randomData(32);
+        entry.softTtl = 8765432L;
+        entry.ttl = 9876543L;
+        putEntry("key", entry).get();
+
+        invalidateEntry("key", false).get();
```
sophiahaim, line 315: Add param:
/* fullExpire= */ false

Same for test below

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,449 @@
...
+        assertThat(getEntry("key1").get(), is(nullValue()));
+
+        // confirm that we can now store entries
+        putEntry("key2", entry).get();
+        assertThatEntriesAreEqual(getEntry("key2").get(), entry);
+    }
+
+    /* Test helpers */
+
+    /** Puts entry into the cache, and returns a CompletableFuture, to let us know when it's done */
```
sophiahaim, line 360: nit: over the line limit
super nit: Add a period on the end (same for #getEntry)

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,449 @@
...
+                fullExpire,
+                new AsyncCache.OnCompleteCallback() {
+                    @Override
+                    public void onComplete() {
+                        remove.complete(null);
+                    }
+                });
+        return remove;
+    }
+
```
sophiahaim, line 415: Optional: Could be worth creating a CacheTestUtil with the methods below since they are used in the DiskBasedCacheTest as well

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +176,97 @@ public void completed(Integer resultLen, Void ignore) {
...
+    @Override
+    public void initialize(OnCompleteCallback callback) {
+        File rootDirectory = mRootDirectorySupplier.get();
+        if (!rootDirectory.exists()) {
+            if (!rootDirectory.mkdirs()) {
+                VolleyLog.e("Unable to create cache dir %s", rootDirectory.getAbsolutePath());
+            }
+            return;
+        }
+        File[] files = rootDirectory.listFiles();
```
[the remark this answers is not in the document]
sphill99: It's my understanding that as long as rootDirectory is a folder and not a file, the output of listFiles() will never be null. But it could be a good check just in case, as we do it in DiskBasedCache.

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sphill99: I thought it'd be better since it's only only one level of indentation, and it'd be more code to use the future. But it definitely is less readable, now that I look at it again. I just changed it.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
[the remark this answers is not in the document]
joebowbeer: @sphill99 According to the method docs, the check is needed:

`Returns null if this abstract pathname does not denote a directory, or if an I/O error occurs.`

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sphill99: That's what the test in DiskBasedCache did, but I think you're right that it is supposed to check the maxsize entry.

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -115,6 +120,41 @@ static CacheHeader readHeader(DiskBasedCache.CountingInputStream is) throws IOEx
...
+            String key = DiskBasedCacheUtility.readString(buffer);
+            String etag = DiskBasedCacheUtility.readString(buffer);
+            long serverDate = buffer.getLong();
+            long lastModified = buffer.getLong();
+            long ttl = buffer.getLong();
+            long softTtl = buffer.getLong();
+            List<Header> allResponseHeaders = DiskBasedCacheUtility.readHeaderList(buffer);
+            return new CacheHeader(
+                    key, etag, serverDate, lastModified, ttl, softTtl, allResponseHeaders);
+        } catch (ExecutionException e) {
```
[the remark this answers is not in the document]
joebowbeer: Be careful about reporting execution exceptions, which are not expected, as ioexceptions.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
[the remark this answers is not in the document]
sphill99: oh thanks! just added it

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,449 @@
...
+        Cache.Entry entry = randomData(MAX_SIZE - getEntrySizeOnDisk("maxsize") - 1);
+        putEntry("maxsize", entry).get();
+
+        assertThatEntriesAreEqual(getEntry("maxsize").get(), entry);
+    }
+
+    @Test
+    public void testTrimAtThreshold() throws ExecutionException, InterruptedException {
+        // Start with the largest possible entry.
+        Cache.Entry entry = randomData(MAX_SIZE - getEntrySizeOnDisk("maxsize") - 1);
```
joebowbeer, line 113: I think largest size is now MAX_SIZE if `=` was removed from check in putEntry?

sphill99: yep, I'll make that change here and in DiskBasedCacheTest

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
joebowbeer: @sophiahaim I agree. Should be maxsize. Slipped through unnoticed in https://github.com/google/volley/pull/196

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -193,20 +277,20 @@ private void initializeIfRootDirectoryDeleted() {
         VolleyLog.d("Re-initializing cache after external clearing.");
         mEntries.clear();
         mTotalSize = 0;
-        initialize();
+        initialize(emptyCallback);
```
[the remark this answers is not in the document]
sphill99: Hmm I see what you're saying. Would you suggest that we wait to return until initialize calls its callback? And would that solve this problem?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +176,97 @@ public void completed(Integer resultLen, Void ignore) {
...
+        for (File file : files) {
+            Path path = file.toPath();
+            AsynchronousFileChannel afc = null;
+            try {
+                afc = AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+                int entrySize = (int) file.length();
+                CacheHeader entry = CacheHeader.readHeader(afc, entrySize);
+                afc.close();
+                entry.size = entrySize;
+                mTotalSize = DiskBasedCacheUtility.putEntry(entry.key, entry, mTotalSize, mEntries);
```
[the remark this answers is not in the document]
sphill99: On my screen it's showing it's just at the line limit. I can put the arguments on new lines, but not sure if that'd be more readable.

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sphill99: Currently, almost all of the statements fall on one line, and I think making this change would make a lot of them multiple line statements and harder to read. With that being said, it is a decent amount of duplicated code, so I think there are benefits to either options.

src/main/java/com/android/volley/toolbox/CacheHeader.java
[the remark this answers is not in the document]
sphill99: Do you suggest adding these exceptions to the method header, then catching them like the IOException in DiskBasedAsyncCache? I don't see how they'd be handled differently, since we still would have to return null on the get call or skip the file in initialize.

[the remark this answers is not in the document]
joebowbeer: It is tricky mixing sync and async code. Java doesn't have a great solution yet AFAIK. We punted on this when we created the Executor framework in java.util.concurrent.

1. Avoid morphing non-IO exceptions into IO exceptions. IOExceptions are checked and have specific causes. In this scenario, there shouldn't be any other checked exceptions, right? Except possibly interrupted exception, but it can be rethrown. Any other exception is probably a programming error and should not be disguised, caught, and then handled by the client's IOException handler.

2. Consider mimicking what would happen if the code executed synchronously. If an unchecked exception is the cause you could attempt to rethrown it in some fashion.

There are a few solutions floating around. Some of them are attempts to avoid checked exceptions altogether, but this is a more noble endeavor.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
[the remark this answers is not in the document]
sophiahaim: Oh, maybe it's just an issue on my screen then – you can ignore this, thanks!

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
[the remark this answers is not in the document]
sophiahaim: Yea, I think we may have to. Otherwise sending the #onComplete callback from the caller doesn't really seem accurate.

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sophiahaim: I'd personally learn towards not having duplicate code, and hopefully the statements are still pretty readable, but it's up to you :)

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -225,4 +312,42 @@ private void deleteFile(File file) {
...
     }
+
+    /** Empty onCompleteCallback when no operation is needed in the callback. */
+    private OnCompleteCallback emptyCallback =
+            new OnCompleteCallback() {
+                @Override
+                public void onComplete() {
+                    // do nothing
+                }
+            };
```
jbir789, line 323: This is a variable, so I'd suggest to move it at the top. Also, given that it's a constant, could you make it `static final`?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -102,13 +114,14 @@ public void failed(Throwable exc, Void ignore) {
         } catch (IOException e) {
             VolleyLog.e(e, "Failed to read file %s", file.getAbsolutePath());
             closeChannel(channel, "IOException");
+            remove(key, emptyCallback);
```
jbir789, line 117: Same here, I think we should wait until the callback returns before moving on

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,450 @@
...
+
+    @Before
+    public void setup() throws IOException {
+        emptyCallback =
+                new AsyncCache.OnCompleteCallback() {
+                    @Override
+                    public void onComplete() {
+                        // do nothing
+                    }
+                };
```
jbir789, line 58: setup won't wait for initialize to be complete if we do this. Could you use a completable future?

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,450 @@
...
+        // Initialize empty cache
+        cache = new DiskBasedAsyncCache(temporaryFolder.getRoot(), MAX_SIZE);
+        cache.initialize(emptyCallback);
+    }
+
+    @After
+    public void teardown() {
+        cache = null;
```
jbir789, line 66: Is this needed? I would think the field variable is re-created between each tests

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
```diff
@@ -0,0 +1,450 @@
...
+        entry.softTtl = 8765432L;
+        entry.etag = "etag";
+        entry.responseHeaders = new HashMap<>();
+        entry.responseHeaders.put("fruit", "banana");
+        entry.responseHeaders.put("color", "yellow");
+
+        putEntry("my-magical-key", entry).get();
+
+        assertThatEntriesAreEqual(getEntry("my-magical-key").get(), entry);
+        assertThat(getEntry("unknown-key").get(), is(nullValue()));
```
jbir789, line 91: can you do: 

```java
assertNull(getEntry(...).get());
```
?

sphill99: Yeah, I kept the same format from DiskBasedCacheTest. But I think doing assertNull makes more sense.

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sphill99: The tests run just fine without it. It was in DiskBasedCacheTest though, so I left it.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: For here, should I use a CompletableFuture to wait until it returns? My hesitation there is that we'd then have to deal with more exceptions, when we're already in the catch block.

sphill99: Actually, I can just call get's callback in the callback instead of doing anything with a future. But in this case, does it necessarily matter? Since the key will be mapped to nothing anyway?

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sphill99: Yep

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I ended up just getting rid of this

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jbir789: I see your point, but it'd feel safer if the get callback only returns once the work is fully complete, including when it returns nothing. This way, the caller knows when the async work is complete (and the callback method is called `onGetComplete`)

sphill99: yeah I agree. I ended up just caliing onGetComplete in the onComplete callback.

src/test/java/com/android/volley/toolbox/DiskBasedAsyncCacheTest.java
sphill99: Went through and changed this in DiskBasedCacheTest and DiskBasedCacheAsyncTest

src/main/java/com/android/volley/toolbox/CacheHeader.java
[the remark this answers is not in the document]
sphill99: I decided to return null instead of throwing it as an IOException, then dealing with it in DBAC. Let me know if there are any issues with this approach.

sphill99:
I believe I've addressed everything in this pull request. Let me know if anything wasn't fully addressed.

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -115,6 +120,40 @@ static CacheHeader readHeader(DiskBasedCache.CountingInputStream is) throws IOEx
...
+    /**
+     * Reads the header from a AsynchronousFileChannel and returns a CacheHeader object.
+     *
+     * @param afc File channel to read from.
+     * @param size Size of the cache header.
+     * @throws IOException if fails to read header
+     */
+    @RequiresApi(api = Build.VERSION_CODES.O)
+    @Nullable
+    static CacheHeader readHeader(final AsynchronousFileChannel afc, int size) throws IOException {
```
joebowbeer, line 132: Several problems here:
1. Declares that it throws IOException but returns null instead? Static analysis (e.g. FindBugs) should flag this?
2. All exceptions are hidden without a log message.
3. Why throw IOE if you're going to return null anyway?
4. What about interrupted exception?

The smallest fix is to move the try-catch to contain only the future.get().
Handle interrupted and execution exception there:
* The cause of the execution exception is very likely to be IOException. I suggest rethrowing the cause wrapped in a new IOException at this point.
* How to handle interrupted exception? Rethrow as IOException is one idea.

Then remove the outer try-catch and let IOException in body throw from method normally.

CompletionHandler here would allow you to compose this operation and postpone these decisions?

sphill99: I'm a little confused, since I thought the original issue was rethrowing the exceptions as IOExceptions? I just updated the code, so that the try-catch only contains future.get(), logs both exceptions so they aren't hidden, and then rethrows as an IOException. Does this fix your concerns?

I had tried to use a completion handler, but the code seemed to be a good bit more complicated.

joebowbeerxealth: Sorry for the run-around. On closer inspection I see that my InterruptedException "suggestion" was incorrect. This InterruptedException happens if the caller to `future.get()` is interrupted while waiting. This is unlikely if the caller is running on the main thread.

One typical way to handle this is to reassert the interrupt so it's not swallowed. I think that's what you should do. (Check out the chapter in Effective Java for more guidance.)

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -115,6 +120,44 @@ static CacheHeader readHeader(DiskBasedCache.CountingInputStream is) throws IOEx
...
+     */
+    @RequiresApi(api = Build.VERSION_CODES.O)
+    static CacheHeader readHeader(final AsynchronousFileChannel afc, int size) throws IOException {
+        final ByteBuffer buffer = ByteBuffer.allocate(size);
+        Future<Integer> fillBuffer = afc.read(buffer, 0);
+        try {
+            fillBuffer.get();
+        } catch (ExecutionException e) {
+            VolleyLog.e(e.toString(), "ExecutionException thrown");
+            throw new IOException(e);
```
joebowbeerxealth, line 138: Yes, this is what I meant.

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -115,6 +120,44 @@ static CacheHeader readHeader(DiskBasedCache.CountingInputStream is) throws IOEx
...
+        } catch (InterruptedException e) {
+            VolleyLog.e(e.toString(), "InterruptedException thrown");
```
joebowbeerxealth, line 140: `caught` not `thrown`?

I think this is a reasonable way to handle interruption.

src/main/java/com/android/volley/toolbox/CacheHeader.java
sphill99: okay, thanks!

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -24,20 +24,49 @@
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
 
-    public interface OnPutCompleteCallback {
-        /** Invoked when the put to the cache is complete. */
-        void onPutComplete();
+    public interface OnCompleteCallback {
```
jpd236, line 27: Maybe we should call this "OnWriteCompleteCallback" to distinguish it from OnGetCompleteCallback. The unifying factor in all of the places which use this callback is that it's an asynchronous write to the cache (where we don't care about the result of the write, just that it's done).

src/main/java/com/android/volley/toolbox/CacheHeader.java
jpd236: Sorry for joining this discussion late!

I think the reason we're running into pain here is because we're doing blocking I/O in a non-blocking API implementation. The aim of the asynchronous API is that we should always return immediately, without blocking on any of the disk operations, including reading of the cache header. Therefore, we shouldn't be making a blocking call to get() inside the asynchronous get method. If we're doing any sort of operation that can throw an InterruptedException in an asynchronous method, that's usually not a good sign.

We should start by reading the header asynchronously, and chain a callback once that read completes to either handle invalid headers or else perform the rest of the read, and then dispatch the user's callback once it completes. This can be done with nested callbacks, or by chaining futures if supported at this API level.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +181,102 @@ public void completed(Integer resultLen, Void ignore) {
...
+            callback.onComplete();
+            return;
+        }
+        for (File file : files) {
+            Path path = file.toPath();
+            AsynchronousFileChannel afc = null;
+            try {
+                afc = AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+                int entrySize = (int) file.length();
+                CacheHeader entry = CacheHeader.readHeader(afc, entrySize);
```
jpd236, line 241: Similarly, here we'll want to kick off asynchronous reads of each file (ideally in parallel), and as the results come in, we should populate our map, rather than doing this effectively sequentially and in a blocking manner. CompletableFuture.allOf would likely help here.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +181,102 @@ public void completed(Integer resultLen, Void ignore) {
...
+        callback.onComplete();
+    }
+
+    /** Invalidates an entry in the cache. */
+    @Override
+    public void invalidate(
+            final String key, final boolean fullExpire, final OnCompleteCallback callback) {
+        Cache.Entry entry = null;
+        try {
+            entry = getFuture(key).get();
```
jpd236, line 259: the get/put should be chained, not interrupting.

src/main/java/com/android/volley/AsyncCache.java
sphill99: Yeah, I think that'd be a good change.

sphill99:
The last commit should have gotten rid of all the blocking code in the API. Now, we read the file all at once, and construct the header from the filled ByteBuffer. Also changed the initialize method, so that we start all reads, then call its callback when they're all finished. Since readHeader is now called in an inner class, I return null instead of throwing an IOException, as the code seemed to be cleaner that way. Let me know what you think about these changes.

joebowbeer:
Why is AsyncCache an abstract class instead of an interface?

src/main/java/com/android/volley/AsyncCache.java
```diff
@@ -24,20 +24,50 @@
      */
     public abstract void get(String key, OnGetCompleteCallback callback);
```
joebowbeer, line 25: Why abstract class?

sphill99: Having it be an abstract class gives us more flexibility, since we can add methods without breaking anything. And the android API guidelines says to favor abstract classes for libraries targeting Java 7, unless you need multiple inheritance.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +174,152 @@ public void completed(Integer resultLen, Void ignore) {
...
+                            public void completed(Integer integer, File ignore) {
+                                CacheHeader entry = CacheHeader.readHeader(buffer);
+                                if (entry != null) {
+                                    closeChannel(afc, "after successful read");
+                                    entry.size = entrySize;
+                                    mTotalSize =
+                                            DiskBasedCacheUtility.putEntry(
+                                                    entry.key, entry, mTotalSize, mEntries);
+                                } else {
+                                    closeChannel(afc, "after failed read");
```
sophiahaim, line 256: Should you call deleteFile here?

sphill99: I didn't since it would be pruned anyway, but maybe it is better to just delete here.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -236,6 +236,15 @@ static String readString(CountingInputStream cis) throws IOException {
         return new String(b, "UTF-8");
     }
 
+    static String readString(ByteBuffer buffer) throws IOException {
```
jpd236, line 239: I think it's a bit strange for this to throw IOException - I think the only thing it can throw is UnsupportedEncodingException in new String, but this wouldn't be thrown by anything else in this method, right?

I wouldn't clean this up in this PR since it's a pattern we're taking from elsewhere, but in a follow-up PR, I would suggest perhaps adding a new method newUtf8String which catches UnsupportedEncodingException and rethrows it as a RuntimeException. UTF-8 should always exist so we shouldn't have to clutter code with handling in case it doesn't. Then callers won't have to deal with IOExceptions here - just the BufferUnderflowException if we run out of buffer.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
```diff
@@ -277,6 +286,21 @@ static void writeHeaderList(@Nullable List<Header> headers, ByteBuffer buffer)
         return result;
     }
 
+    static List<Header> readHeaderList(ByteBuffer buffer) throws IOException {
+        int size = buffer.getInt();
+        if (size < 0) {
+            throw new IOException("readHeaderList size=" + size);
+        }
+        List<Header> result =
+                (size == 0) ? Collections.<Header>emptyList() : new ArrayList<Header>();
```
jpd236, line 295: I would keep this simple and just do new ArrayList<Header>(size). Optimizing to return emptyList() is actually a sort of anti-pattern here - it means that this method sometimes returns an immutable object (when empty) and sometimes doesn't (when non-empty), which can lead to bugs.

src/main/java/com/android/volley/toolbox/CacheHeader.java
```diff
@@ -115,6 +115,36 @@ static CacheHeader readHeader(DiskBasedCache.CountingInputStream is) throws IOEx
...
+     * Reads the header from a AsynchronousFileChannel and returns a CacheHeader object.
+     *
+     * @param buffer Buffer to get header info from.
+     * @throws IOException if fails to read header
+     */
+    @Nullable
+    static CacheHeader readHeader(final ByteBuffer buffer) {
+        try {
+            buffer.flip();
+            int magic = buffer.getInt();
```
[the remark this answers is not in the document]
sphill99: Just to check, should I catch all runtime exceptions or just a BufferUnderflowException?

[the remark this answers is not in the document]
jpd236: Just BufferUnderflow - anything else would be an unexpected/developer error, even in the face of invalid data. (We should also make sure we handle invalid UTF-8 data when reading strings in case that results in a runtime exception rather than just garbage data, though).

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -70,45 +75,53 @@ public void get(String key, final OnGetCompleteCallback callback) {
...
                                 VolleyLog.e(
                                         "File changed while reading: %s", file.getAbsolutePath());
                                 callback.onGetComplete(null);
-                                return;
                             }
-                            byte[] data = buffer.array();
-                            callback.onGetComplete(entry.toCacheEntry(data));
+                            if (!readHeaderError(key, entryOnDisk, callback, file)) {
+                                int pos = buffer.position();
+                                byte[] data = Arrays.copyOfRange(buffer.array(), pos, size);
```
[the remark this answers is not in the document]
sphill99: yeah, that makes that a lot cleaner.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -70,45 +75,53 @@ public void get(String key, final OnGetCompleteCallback callback) {
...
-                    /* position= */ headerSize,
+                    /* position= */ 0,
                     /* attachment= */ null,
                     new CompletionHandler<Integer, Void>() {
                         @Override
                         public void completed(Integer result, Void v) {
                             closeChannel(afc, "completed read");
-                            // if the file size changes, return null
+                            CacheHeader entryOnDisk = CacheHeader.readHeader(buffer);
                             if (size != result) {
```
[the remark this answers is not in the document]
sphill99: yeah, that's what I was struggling with, since they all are slightly different. Would the best default be to remove as well as delete for all three? Or not delete in any case, and let pruning take care of that?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +174,152 @@ public void completed(Integer resultLen, Void ignore) {
...
+                closeChannel(channel, "IOException in initialize");
+                deleteFile(file);
             }
-            callback.onPutComplete();
+            reads.add(fileRead);
         }
+        CompletableFuture<Void> voidCompletableFuture =
+                CompletableFuture.allOf(reads.toArray(new CompletableFuture[0]));
+
+        voidCompletableFuture.thenRun(
```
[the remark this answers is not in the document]
sphill99: Yes, the initial setup would never happen unless the callback is invoked.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -223,6 +360,34 @@ private void deleteFile(File file) {
...
+    }
+
+    /** Attempts to create the root directory, logging if it was unable to. */
+    private void createCacheDirectory(File rootDirectory) {
+        if (!rootDirectory.mkdirs()) {
+            VolleyLog.e("Unable to create cache dir %s", rootDirectory.getAbsolutePath());
+        }
+    }
+
+    /** Closes the file channel, removes the key, calls the callback. */
```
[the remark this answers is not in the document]
sphill99: that was something that I had before and changed. Also changing the name.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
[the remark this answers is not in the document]
jpd236: my inclination is to fully prune an entry that we're unable to read regardless of the specific error that occurred.

src/main/java/com/android/volley/toolbox/DiskBasedCacheUtility.java
sphill99: Also, since all these methods require a higher API, we can replace "UTF-8" with StandardCharsets.UTF-8 and that doesn't throw any exception. It wouldn't be the same as code for DiskBasedCache, but might be better since there'd be no error handling needed.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -161,28 +174,152 @@ public void completed(Integer resultLen, Void ignore) {
...
+            try {
+                channel = AsynchronousFileChannel.open(path, StandardOpenOption.READ);
+                final AsynchronousFileChannel afc = channel;
+                afc.read(
+                        buffer,
+                        0,
+                        file,
+                        new CompletionHandler<Integer, File>() {
+                            @Override
+                            public void completed(Integer integer, File ignore) {
```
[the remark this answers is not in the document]
sphill99: yeah, I was also thinking about this. currently we allocate the entire size of the file for the ByteBuffer, but that's a huge waste. Not sure how to optimize without knowing the size of the header before reading it.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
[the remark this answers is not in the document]
sphill99: okay, that makes sense. Should we still log the specific issues in each case? Or should we group them all together, and just log that the file couldn't be read?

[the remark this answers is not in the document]
jpd236: Ideally we'd log the specific issue as that might help diagnose the nature in which some corruption was happening (or perhaps a bug in one of our checks).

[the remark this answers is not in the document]
sphill99: okay, I log the error then call a helper that removes the key, deletes the file, and calls the callback in each of these three cases. Should that be true for if the read fails (goes into failed function of completion handler)?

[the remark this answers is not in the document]
jpd236: Seems the same as any other kind of IO failure to me :)

sphill99:
Every comment should have been addressed by now. The only two that I didn't implement were optimizing initialize, so that it doesn't read the entire file just to get the each, and handling UnsupportedEncodingException in different read methods.

joebowbeer:
If you're only using UTF-8 character encoding, UnsupportedEncodingException should be treated as an unexpected/fatal error because this encoding is the default on Androud, and is always available.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -70,45 +74,57 @@ public void get(String key, final OnGetCompleteCallback callback) {
...
                     new CompletionHandler<Integer, Void>() {
                         @Override
                         public void completed(Integer result, Void v) {
                             closeChannel(afc, "completed read");
-                            // if the file size changes, return null
                             if (size != result) {
                                 VolleyLog.e(
                                         "File changed while reading: %s", file.getAbsolutePath());
-                                callback.onGetComplete(null);
+                                removeDeleteAndCall(key, callback, file);
```
jpd236, line 90: nit: I think this method name is unclear (remove what? delete what? call what?).

I think it's arguably implied that removing the file should imply it gets deleted from the map, so calling this something like "deleteFileAndInvokeCallback" would be a bit more readable.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -70,45 +74,57 @@ public void get(String key, final OnGetCompleteCallback callback) {
...
+                            } else {
+                                byte[] data = new byte[buffer.remaining()];
+                                buffer.get(data);
+                                callback.onGetComplete(entry.toCacheEntry(data));
+                            }
                         }
 
                         @Override
                         public void failed(Throwable exc, Void ignore) {
+                            closeChannel(afc, "completed read");
```
jpd236, line 113: not sure why this was reverted from "failed" to "completed"

joebowbeer:
I'd like to see a high level explanation for how this async implementation works. Aside from afc operation, by the way, is everything running on the main thread?

My concern: I see final file variables that are assigned on the main thread, then some async stuff happens, which might prune them, replace them, or both, then the async completion handler is invoked and it expects to be able to write or delete that same file. How is this enforced?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
jpd236: was this meant to imply you'd change it to delete? since it doesn't seem to be done. Since DiskBasedCache#initialize deletes the file in the event of any IO exception (which includes invalid headers), I think we should do the same here.

jpd236:
> If you're only using UTF-8 character encoding, UnsupportedEncodingException should be treated as an unexpected/fatal error because this encoding is the default on Androud, and is always available.

(Yes, agreed - but we can defer this to follow-up cleanup since it impacts existing code as well)

> I'd like to see a high level explanation for how this async implementation works. Aside from afc operation, by the way, is everything running on the main thread?
> 
> My concern: I see final file variables that are assigned on the main thread, then some async stuff happens, which might prune them, replace them, or both, then the async completion handler is invoked and it expects to be able to write or delete that same file. How is this enforced?

As the PR title notes, there are definitely thread safety issues we need to resolve before this would be ready for use. In terms of a broader design, we have an internal document we've been working off of. It's unfortunately not suitable for external sharing at the moment, but we're sharing details in the context of each PR.

joebowbeer:
Especially if you have a pool of threads serving afc, and not, say, a single worker, and don't serialize requests in some other way, then I don't see how you can assume that a file handle will persist across an async invocation. That file could already be deleted, along with the entire cache, or a newer version of that file may already have been written.

jpd236:
At a high level, we plan to have ReadWriteLocks per cache entry to ensure that while a write is happening, reads or other writes cannot happen to the same entry, while still permitting fully parallel access for different files.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I changed it to that; it's definitely more descriptive.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I don't know how I did that. Just changed it back.

joebowbeer:
I'm interested to know how prune will be done. I imagine it may move back to after-write and run if needed whenever all the locks are released.

PR #356, opened 2020-07-31
AsyncStack class added, as well as Cronet implementation. Limited unit testing

sphill99:
I uploaded the AsyncStack abstract class, as well as the Cronet extension of it. I also added a constructor for HttpResponse that takes in a byte[] directly. I added tests for the one helper, but most things I was not able to add in unit tests at this time. I did test different calls to executeResponse in a sample test app and all those were working.

Few open questions:
- in the middle of reading, should we be checking the status code and potentially canceling early? Or should all of that be done in the network? 
- Is there anything more I need to set in the UrlRequest.Builder? I currently set request headers and the type, as well as adding the request body, if it exists.

### DIFF
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -1,0 +1,248 @@
+package com.android.volley.cronet;
+
+import android.content.Context;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
+import com.android.volley.AuthFailureError;
+import com.android.volley.Header;
+import com.android.volley.Request;
+import com.android.volley.toolbox.AsyncHttpStack;
+import com.android.volley.toolbox.ByteArrayPool;
+import com.android.volley.toolbox.HttpResponse;
+import com.android.volley.toolbox.PoolingByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
+import java.nio.channels.WritableByteChannel;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import org.chromium.net.CronetEngine;
+import org.chromium.net.CronetException;
+import org.chromium.net.UploadDataProvider;
+import org.chromium.net.UploadDataProviders;
+import org.chromium.net.UrlRequest;
+import org.chromium.net.UrlRequest.Callback;
+import org.chromium.net.UrlResponseInfo;
+
+public class CronetHttpStack extends AsyncHttpStack {
+
+    private final CronetEngine mCronetEngine;
+    private final Executor mCallbackExecutor;
+    private final ByteArrayPool mPool;
+
+    public CronetHttpStack(Context context, Executor callbackExecutor, ByteArrayPool pool) {
+        mCronetEngine = new CronetEngine.Builder(context).build();
+        mCallbackExecutor = callbackExecutor;
+        mPool = pool;
+    }
+
+    public CronetHttpStack(Context context) {
+        this(context, (ByteArrayPool) null);
+    }
+
+    public CronetHttpStack(Context context, Executor callbackExecutor) {
+        this(context, callbackExecutor, null);
+    }
+
+    public CronetHttpStack(Context context, ByteArrayPool pool) {
+        this(
+                context,
+                new ThreadPoolExecutor(
+                        0,
+                        1,
+                        60,
+                        TimeUnit.SECONDS,
+                        new LinkedBlockingQueue<Runnable>(),
+                        new ThreadFactory() {
+                            @Override
+                            public Thread newThread(@NonNull Runnable runnable) {
+                                Thread t = Executors.defaultThreadFactory().newThread(runnable);
+                                t.setName("Volley-CronetCallbackThread");
+                                return t;
+                            }
+                        }),
+                pool);
+    }
+
+    @Override
+    public void executeRequest(
+            final Request<?> request,
+            final Map<String, String> additionalHeaders,
+            final OnRequestComplete callback) {
+        final PoolingByteArrayOutputStream bytesReceived = new PoolingByteArrayOutputStream(mPool);
+        final WritableByteChannel receiveChannel = Channels.newChannel(bytesReceived);
+        final Callback urlCallback =
+                new Callback() {
+                    @Override
+                    public void onRedirectReceived(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo, String s) {
+                        urlRequest.followRedirect();
+                    }
+
+                    @Override
+                    public void onResponseStarted(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo) {
+                        int httpStatusCode = urlResponseInfo.getHttpStatusCode();
+                        // what checks should we be making of the status code, if any?
+                        if (mPool != null) {
+                            urlRequest.read(ByteBuffer.wrap(mPool.getBuf(102400)));
+                        } else {
+                            urlRequest.read(ByteBuffer.allocateDirect(102400));
+                        }
+                    }
+
+                    @Override
+                    public void onReadCompleted(
+                            UrlRequest urlRequest,
+                            UrlResponseInfo urlResponseInfo,
+                            ByteBuffer byteBuffer) {
+                        byteBuffer.flip();
+                        try {
+                            receiveChannel.write(byteBuffer);
+                            byteBuffer.clear();
+                            urlRequest.read(byteBuffer);
+                        } catch (IOException e) {
+                            callback.onError(e);
+                            urlRequest.cancel();
+                        }
+                    }
+
+                    @Override
+                    public void onSucceeded(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo) {
+                        List<Header> headers = getHeaders(urlResponseInfo.getAllHeadersAsList());
+                        HttpResponse response =
+                                new HttpResponse(
+                                        urlResponseInfo.getHttpStatusCode(),
+                                        headers,
+                                        bytesReceived.toByteArray());
+                        callback.onSuccess(response);
+                    }
+
+                    @Override
+                    public void onFailed(
+                            UrlRequest urlRequest,
+                            UrlResponseInfo urlResponseInfo,
+                            CronetException e) {
+                        if (urlResponseInfo != null) {
+                            List<Header> headers =
+                                    getHeaders(urlResponseInfo.getAllHeadersAsList());
+                            HttpResponse response =
+                                    new HttpResponse(
+                                            urlResponseInfo.getHttpStatusCode(),
+                                            headers,
+                                            bytesReceived.toByteArray());
+                            callback.onSuccess(response);
+                        } else {
+                            callback.onError(e);
+                        }
+                    }
+                };
+        UrlRequest.Builder builder =
+                mCronetEngine
+                        .newUrlRequestBuilder(request.getUrl(), urlCallback, mCallbackExecutor)
+                        .allowDirectExecutor();
+        try {
+            setHttpMethod(request, builder);
+            setRequestHeaders(request, additionalHeaders, builder);
+            UrlRequest urlRequest = builder.build();
+            urlRequest.start();
+        } catch (AuthFailureError authFailureError) {
+            callback.onAuthError(authFailureError);
+        }
+    }
+
+    @VisibleForTesting
+    public static List<Header> getHeaders(List<Map.Entry<String, String>> headersList) {
+        List<Header> headers = new ArrayList<>();
+        for (Map.Entry<String, String> header : headersList) {
+            headers.add(new Header(header.getKey(), header.getValue()));
+        }
+        return headers;
+    }
+
+    /** Sets the connection parameters for the UrlRequest */
+    private void setHttpMethod(Request<?> request, UrlRequest.Builder builder)
+            throws AuthFailureError {
+        switch (request.getMethod()) {
+            case Request.Method.DEPRECATED_GET_OR_POST:
+                // This is the deprecated way that needs to be handled for backwards compatibility.
+                // If the request's post body is null, then the assumption is that the request is
+                // GET.  Otherwise, it is assumed that the request is a POST.
+                byte[] postBody = request.getPostBody();
+                if (postBody != null) {
+                    builder.setHttpMethod("POST");
+                    addBodyIfExists(postBody, builder);
+                }
+                break;
+            case Request.Method.GET:
+                // Not necessary to set the request method because connection defaults to GET but
+                // being explicit here.
+                builder.setHttpMethod("GET");
+                break;
+            case Request.Method.DELETE:
+                builder.setHttpMethod("DELETE");
+                break;
+            case Request.Method.POST:
+                builder.setHttpMethod("POST");
+                addBodyIfExists(request.getBody(), builder);
+                break;
+            case Request.Method.PUT:
+                builder.setHttpMethod("PUT");
+                addBodyIfExists(request.getBody(), builder);
+                break;
+            case Request.Method.HEAD:
+                builder.setHttpMethod("HEAD");
+                break;
+            case Request.Method.OPTIONS:
+                builder.setHttpMethod("OPTIONS");
+                break;
+            case Request.Method.TRACE:
+                builder.setHttpMethod("TRACE");
+                break;
+            case Request.Method.PATCH:
+                builder.setHttpMethod("PATCH");
+                addBodyIfExists(request.getBody(), builder);
+                break;
+            default:
+                throw new IllegalStateException("Unknown method type.");
+        }
+    }
+
+    /**
+     * Sets the request headers for the UrlRequest.
+     *
+     * @param request to get the headers from
+     * @param additionalHeaders for the UrlRequest
+     * @param builder that we are adding the request headers to
+     * @throws AuthFailureError is thrown if Request#getHeaders throws ones
+     */
+    private void setRequestHeaders(
+            Request<?> request, Map<String, String> additionalHeaders, UrlRequest.Builder builder)
+            throws AuthFailureError {
+        HashMap<String, String> map = new HashMap<>();
+        map.putAll(additionalHeaders);
+        // Request.getHeaders() takes precedence over the given additional (cache) headers).
+        map.putAll(request.getHeaders());
+        for (Map.Entry<String, String> header : map.entrySet()) {
+            builder.addHeader(header.getKey(), header.getValue());
+        }
+    }
+
+    /** Sets the UploadDataProvider of the UrlRequest.Builder */
+    private void addBodyIfExists(@Nullable byte[] body, UrlRequest.Builder builder) {
+        if (body != null) {
+            UploadDataProvider dataProvider = UploadDataProviders.create(body);
+            builder.setUploadDataProvider(dataProvider, mCallbackExecutor);
+        }
+    }
+}
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -6,7 +6,6 @@ import androidx.annotation.VisibleForTesting;
 import com.android.volley.AuthFailureError;
 import com.android.volley.Header;
 import com.android.volley.Request;
-import com.android.volley.VolleyLog;
 import com.android.volley.toolbox.AsyncHttpStack;
 import com.android.volley.toolbox.ByteArrayPool;
 import com.android.volley.toolbox.HttpResponse;
@@ -81,7 +80,12 @@ public class CronetHttpStack extends AsyncHttpStack {
             Executor callbackExecutor,
             ExecutorService blockingExecutor,
             UrlRewriter rewriter) {
-        this(context, callbackExecutor, blockingExecutor, new ByteArrayPool(DEFAULT_POOL_SIZE), rewriter);
+        this(
+                context,
+                callbackExecutor,
+                blockingExecutor,
+                new ByteArrayPool(DEFAULT_POOL_SIZE),
+                rewriter);
     }
 
     @Override
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -41,53 +41,12 @@ public class CronetHttpStack extends AsyncHttpStack {
     private final ByteArrayPool mPool;
     private final UrlRewriter mUrlRewriter;
 
-    public CronetHttpStack(
-            Context context,
-            Executor callbackExecutor,
-            ExecutorService blockingExecutor,
-            ByteArrayPool pool,
-            UrlRewriter rewriter) {
-        mCronetEngine = new CronetEngine.Builder(context).build();
-        mCallbackExecutor = callbackExecutor;
-        mBlockingExecutor = blockingExecutor;
-        mPool = pool;
-        mUrlRewriter = rewriter;
-    }
-
-    public CronetHttpStack(
-            Context context,
-            Executor callbackExecutor,
-            ExecutorService blockingExecutor,
-            ByteArrayPool pool) {
-        this(
-                context,
-                callbackExecutor,
-                blockingExecutor,
-                pool,
-                new UrlRewriter() {
-                    @Override
-                    public String rewriteUrl(String originalUrl) {
-                        return originalUrl;
-                    }
-                });
-    }
-
-    public CronetHttpStack(
-            Context context, ExecutorService blockingExecutor, Executor callbackExecutor) {
-        this(context, callbackExecutor, blockingExecutor, new ByteArrayPool(DEFAULT_POOL_SIZE));
-    }
-
-    public CronetHttpStack(
-            Context context,
-            Executor callbackExecutor,
-            ExecutorService blockingExecutor,
-            UrlRewriter rewriter) {
-        this(
-                context,
-                callbackExecutor,
-                blockingExecutor,
-                new ByteArrayPool(DEFAULT_POOL_SIZE),
-                rewriter);
+    private CronetHttpStack(Builder builder) {
+        mCronetEngine = builder.mCronetEngine;
+        mCallbackExecutor = builder.mCallbackExecutor;
+        mBlockingExecutor = builder.mBlockingExecutor;
+        mPool = builder.mPool;
+        mUrlRewriter = builder.mUrlRewriter;
     }
 
     @Override
@@ -95,10 +54,11 @@ public class CronetHttpStack extends AsyncHttpStack {
             final Request<?> request,
             final Map<String, String> additionalHeaders,
             final OnRequestComplete callback) {
-        final PoolingByteArrayOutputStream bytesReceived = new PoolingByteArrayOutputStream(mPool);
-        final WritableByteChannel receiveChannel = Channels.newChannel(bytesReceived);
         final Callback urlCallback =
                 new Callback() {
+                    PoolingByteArrayOutputStream bytesReceived = null;
+                    WritableByteChannel receiveChannel = null;
+
                     @Override
                     public void onRedirectReceived(
                             UrlRequest urlRequest, UrlResponseInfo urlResponseInfo, String s) {
@@ -108,8 +68,11 @@ public class CronetHttpStack extends AsyncHttpStack {
                     @Override
                     public void onResponseStarted(
                             UrlRequest urlRequest, UrlResponseInfo urlResponseInfo) {
-                        int size = getContentLength(urlResponseInfo);
-                        urlRequest.read(ByteBuffer.allocateDirect(size));
+                        bytesReceived =
+                                new PoolingByteArrayOutputStream(
+                                        mPool, getContentLength(urlResponseInfo));
+                        receiveChannel = Channels.newChannel(bytesReceived);
+                        urlRequest.read(ByteBuffer.allocateDirect(1024));
                     }
 
                     @Override
@@ -274,11 +237,64 @@ public class CronetHttpStack extends AsyncHttpStack {
     }
 
     private int getContentLength(UrlResponseInfo urlResponseInfo) {
-        List<String> content = urlResponseInfo.getAllHeaders().get("content-length");
+        List<String> content = urlResponseInfo.getAllHeaders().get("Content-Length");
         if (content == null) {
             return 1024;
         } else {
             return Integer.parseInt(content.get(0));
+        }
+    }
+
+    /**
+     * Builder is used to build an instance of {@link CronetHttpStack} from values configured by the
+     * setters.
+     */
+    public static class Builder {
+        private static final int DEFAULT_POOL_SIZE = 4096;
+        private CronetEngine mCronetEngine;
+        private Executor mCallbackExecutor;
+        private ExecutorService mBlockingExecutor;
+        private ByteArrayPool mPool;
+        private UrlRewriter mUrlRewriter;
+
+        public Builder(
+                Context context, ExecutorService blockingExecutor, Executor callbackExecutor) {
+            mCronetEngine = new CronetEngine.Builder(context).build();
+            mCallbackExecutor = callbackExecutor;
+            mBlockingExecutor = blockingExecutor;
+            mPool = new ByteArrayPool(DEFAULT_POOL_SIZE);
+            mUrlRewriter =
+                    new UrlRewriter() {
+                        @Override
+                        public String rewriteUrl(String originalUrl) {
+                            return originalUrl;
+                        }
+                    };
+        }
+
+        /**
+         * Sets the CronetEngine to be used. Would recommend disabling the HTTP cache, as it is
+         * redundant with Volley's.
+         */
+        public Builder setCronetEngine(CronetEngine engine) {
+            mCronetEngine = engine;
+            return this;
+        }
+
+        /** Sets the ByteArrayPool to be used. */
+        public Builder setPool(ByteArrayPool pool) {
+            mPool = pool;
+            return this;
+        }
+
+        /** Sets the UrlRewriter to be used. */
+        public Builder setUrlRewriter(UrlRewriter urlRewriter) {
+            mUrlRewriter = urlRewriter;
+            return this;
+        }
+
+        public CronetHttpStack build() {
+            return new CronetHttpStack(this);
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.concurrent.TimeUnit;
+import org.chromium.net.CronetEngine;
+import org.chromium.net.CronetException;
+import org.chromium.net.UploadDataProvider;
+import org.chromium.net.UploadDataProviders;
+import org.chromium.net.UrlRequest;
+import org.chromium.net.UrlRequest.Callback;
+import org.chromium.net.UrlResponseInfo;
+
+public class CronetHttpStack extends AsyncHttpStack {
```
jpd236, line 36: In terms of what more you need to do here - I'd go line by line through HurlStack and HttpClientStack and make sure you cover any functionality that makes sense here. At a glance, I see:

- support for a UrlRewriter interface - this is only in HurlStack, but effectively possible in HttpClientStack as well by providing a custom HttpClient.
- handling request timeouts

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+    public CronetHttpStack(Context context, ByteArrayPool pool) {
+        this(
+                context,
+                new ThreadPoolExecutor(
+                        0,
+                        1,
+                        60,
+                        TimeUnit.SECONDS,
+                        new LinkedBlockingQueue<Runnable>(),
+                        new ThreadFactory() {
```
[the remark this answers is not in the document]
sphill99: No, that makes sense to me. I originally didn't have it take in an executor, so I just built on top of this, but I think we can take it out.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+                        urlRequest.followRedirect();
+                    }
+
+                    @Override
+                    public void onResponseStarted(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo) {
+                        int httpStatusCode = urlResponseInfo.getHttpStatusCode();
+                        // what checks should we be making of the status code, if any?
+                        if (mPool != null) {
+                            urlRequest.read(ByteBuffer.wrap(mPool.getBuf(102400)));
```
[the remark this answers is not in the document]
sphill99: I wasn't really sure what to go with there. I think that number was in the prototype, so I kept it. And from my understanding of reading with ByteBuffers, it's generally faster to allocate more than you need than to repeatedly write to the channel and clear. But also this is a waste if it is way too large.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+                        }),
+                pool);
+    }
+
+    @Override
+    public void executeRequest(
+            final Request<?> request,
+            final Map<String, String> additionalHeaders,
+            final OnRequestComplete callback) {
+        final PoolingByteArrayOutputStream bytesReceived = new PoolingByteArrayOutputStream(mPool);
```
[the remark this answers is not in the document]
sphill99: Oh you're right. I forgot to add that check when making this a PoolingByteArrayOuputStream. And yeah, I think that's much cleaner than handling both cases throughout. And then it also gets rid of potential for mistakes like this one.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+        switch (request.getMethod()) {
+            case Request.Method.DEPRECATED_GET_OR_POST:
+                // This is the deprecated way that needs to be handled for backwards compatibility.
+                // If the request's post body is null, then the assumption is that the request is
+                // GET.  Otherwise, it is assumed that the request is a POST.
+                byte[] postBody = request.getPostBody();
+                if (postBody != null) {
+                    builder.setHttpMethod("POST");
+                    addBodyIfExists(postBody, builder);
+                }
```
sophiahaim, line 185: should this have an else with builder.setHttpMethod("GET")? This is what it sounds like should happen from your comment above. Or do you not have to because it's the default? Either way, I think you should be explicit – either set it to "GET" (like below) or add to the comment saying that's the default

sphill99: It is the default, but I'll be explicit and set it.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+                    public void onRedirectReceived(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo, String s) {
+                        urlRequest.followRedirect();
+                    }
+
+                    @Override
+                    public void onResponseStarted(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo) {
+                        int httpStatusCode = urlResponseInfo.getHttpStatusCode();
+                        // what checks should we be making of the status code, if any?
```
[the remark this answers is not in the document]
sphill99: From my understanding, onFailed is called if there isn't a valid response code. By make sure responses are handled appropriately, do you just mean to skip reading if there is no body?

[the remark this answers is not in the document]
jpd236: I mostly mean "identically to how the other stacks do it", to the extent that it make sense. That said, I'd tend to err on the side of simplicity. I can't find any code history for why HurlStack explicitly handles things like 204s rather than just passing them along to the caller; if a server wants to return a 204 with content then that's their problem IMO (if Cronet itself even permits it). So maybe start simpler and we can add the explicit response code handling if we need it.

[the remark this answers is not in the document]
sphill99: sounds good

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,248 @@
...
+                            CronetException e) {
+                        if (urlResponseInfo != null) {
+                            List<Header> headers =
+                                    getHeaders(urlResponseInfo.getAllHeadersAsList());
+                            HttpResponse response =
+                                    new HttpResponse(
+                                            urlResponseInfo.getHttpStatusCode(),
+                                            headers,
+                                            bytesReceived.toByteArray());
+                            callback.onSuccess(response);
```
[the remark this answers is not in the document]
sphill99: yeah, I think you're right. I was thinking that any information should be sent over, but if it's only partial or has other errors, it's probably better to call onError. And I think regardless of the response code, as long as the read is successful, it will end up in onSucceeded, not onFailed.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: I added support for UrlRewriter interface. As we discussed, Cronet doesn't support passing in request timeouts, so we will hold off on that for now. 

I also noticed that the requests all have a Priority number, and we can set priority on the UrlRequest.Builder. Should we set the priority here, or are those two different things?

sphill99:
I believe I either made changes / responded to all comments. I know one thing that's still open is the size to allocate for the ByteBuffer. Similarly, since now we make a default pool if one isn't passed in, should the default size of that be the same as in the BasicNetwork?

jpd236:
Using the same default size as in BasicNetwork seems reasonable as a starting point since the pools are used for the same purposes, as I understand it.

Regarding the buffer size to allocate - can we use the ContentLength response header, at least in some cases? It seems like BasicNetwork is doing this to allocate a byte[] of exactly the right size, though I'm surprised since it seems like we assume the header is always present there, and I don't think it's guaranteed... maybe just long-standing behavior.

sphill99:
I thought we only have the ContentLength once we finish the request? It does look like I can use UrlResponseInfo#getReceivedByteCount() though, which should give us what we need. I just tested it, and that seems to work well. Also, realized I can't wrap byte[] from the pool for these ByteBuffers, since they need to be direct.

jpd236:
The headers come before the body of the response. So they should in theory be available before we start reading the response body. I'm not sure about getReceivedByteCount - would be nice to find some citation indicating that using this to determine how big a buffer to allocate is a good idea. Why do the buffers need to be direct?

sphill99:
The buffers need to be direct, since they access their memory directly. And okay, I understand now. I thought you were saying the ContentLength from the http response, not from the response headers. Looking further into it, getReceivedByteCount would only return the bytes received from the headers. I think what you said should work though. I'm just testing it now to make sure.

joebowbeer:
I notice that Cronet itself has a [configurable cache](https://www.chromium.org/developers/design-documents/network-stack/http-cache).

Example:

`enableHttpCache(CronetEngine.Builder.HTTP_CACHE_IN_MEMORY, 100 * 1024)`

Types:

```
import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_DISABLED;
import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_DISK;
import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_DISK_NO_HTTP;
import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_IN_MEMORY;
```

And a new version of the [disk cache](https://www.chromium.org/developers/design-documents/network-stack/disk-cache) is being developed?

https://www.chromium.org/developers/design-documents/network-stack/disk-cache/disk-cache-v3

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -0,0 +1,93 @@
...
+        Object response = entry.get();
+        if (response instanceof HttpResponse) {
+            return (HttpResponse) response;
+        } else if (response instanceof IOException) {
+            throw (IOException) response;
+        } else if (response instanceof AuthFailureError) {
+            throw (AuthFailureError) response;
+        } else {
+            VolleyLog.d("Response was in an unexpected form");
+            return null;
```
sophiahaim, line 90: Make the method @Nullable since you're returning null in this case, and javadoc the return behavior

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.concurrent.Future;
+import org.chromium.net.CronetEngine;
+import org.chromium.net.CronetException;
+import org.chromium.net.UploadDataProvider;
+import org.chromium.net.UploadDataProviders;
+import org.chromium.net.UrlRequest;
+import org.chromium.net.UrlRequest.Callback;
+import org.chromium.net.UrlResponseInfo;
+
+public class CronetHttpStack extends AsyncHttpStack {
```
sophiahaim, line 33: Add javadoc for this class

src/main/java/com/android/volley/toolbox/HttpResponse.java
```diff
@@ -27,6 +28,7 @@
     private final List<Header> mHeaders;
     private final int mContentLength;
     private final InputStream mContent;
+    private final byte[] mContentBytes;
```
sophiahaim, line 31: @Nullable for this and mContent since it can now be null

src/main/java/com/android/volley/toolbox/HttpResponse.java
```diff
@@ -70,11 +89,28 @@ public final int getContentLength() {
         return mContentLength;
     }
 
+    /**
+     * If a byte[] was already provided by an HTTP stack that natively supports returning one, this
+     * method will return that byte[] as an optimization over copying the bytes from an input
+     * stream. It may return null, even if the response has content, as long as mContent is
```
sophiahaim, line 95: Make method @Nullable. Same for #getContent. I realize that one already could be null, so it may have just been an oversight previously

src/main/java/com/android/volley/toolbox/HurlStack.java
```diff
@@ -41,13 +41,7 @@
...
 
     /** An interface for transforming URLs before use. */
-    public interface UrlRewriter {
-        /**
-         * Returns a URL to use instead of the provided one, or null to indicate this URL should not
-         * be used at all.
-         */
-        String rewriteUrl(String originalUrl);
-    }
+    public interface UrlRewriter extends com.android.volley.toolbox.UrlRewriter {}
```
sophiahaim, line 44: Just for my own understanding, what is this for? It extends the toolbox.UrlRewriter, but doesn't seem to add any methods

sphill99: I didn't want to break the API, since HurlStack could take in a HurlStack.UrlRewriter.

jpd236:
> I notice that Cronet itself has a [configurable cache](https://www.chromium.org/developers/design-documents/network-stack/http-cache).
> 
> Example:
> 
> `enableHttpCache(CronetEngine.Builder.HTTP_CACHE_IN_MEMORY, 100 * 1024)`
> 
> Types:
> 
> ```
> import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_DISABLED;
> import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_DISK;
> import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_DISK_NO_HTTP;
> import static org.chromium.net.CronetEngine.Builder.HTTP_CACHE_IN_MEMORY;
> ```
> 
> And a new version of the [disk cache](https://www.chromium.org/developers/design-documents/network-stack/disk-cache) is being developed?
> 
> https://www.chromium.org/developers/design-documents/network-stack/disk-cache/disk-cache-v3

This is a good point. I don't think we'd want Cronet's HTTP cache because that's already handled by Volley. That said, there are other things that can be cached like QUIC server information which we don't/can't handle ourselves.

In general, I think we should handle this sort of variance by allowing the caller to provide their own CronetEngine instance and configure it however they want, while creating a reasonable default CronetEngine for callers that don't care. For our default, we should probably just disable the cache explicitly.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,284 @@
...
+
+    private static final int DEFAULT_POOL_SIZE = 4096;
+    private final CronetEngine mCronetEngine;
+    private final Executor mCallbackExecutor;
+    private final ExecutorService mBlockingExecutor;
+    private final ByteArrayPool mPool;
+    private final UrlRewriter mUrlRewriter;
+
+    public CronetHttpStack(
```
jpd236, line 44: Let's also allow callers to provide their own CronetEngine, but add Javadoc so we can suggest reasonable parameters, like not enabling the HTTP cache since it's redundant with Volley's cache.

We can continue to have a constructor which doesn't take a CronetEngine and creates a reasonable default, though.

We might also want to consider just having one private master constructor and adding a Builder class to construct stacks. We haven't done this elsewhere in Volley yet, but given the number of parameters and constructors we're getting here, and since this is a brand new class anyway, I think it might be a reasonable option.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+
+    /** Sets the priority of this request. */
+    private void setPriority(Request<?> request, UrlRequest.Builder builder) {
+        builder.setPriority(request.getPriority().ordinal());
+    }
+
+    private int getContentLength(UrlResponseInfo urlResponseInfo) {
+        List<String> content = urlResponseInfo.getAllHeaders().get("content-length");
```
jpd236, line 277: The capitalization should be "Content-Length", and it doesn't appear to me that this is a case-insensitive map (or guaranteed to be one), though most other places that look for headers appear to do so in a case-insensitive fashion:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length
https://developer.android.com/guide/topics/connectivity/cronet/reference/org/chromium/net/UrlResponseInfo.html#getAllHeaders()

is it?

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,284 @@
...
+            builder.setUploadDataProvider(dataProvider, mCallbackExecutor);
+        }
+    }
+
+    /** Sets the priority of this request. */
+    private void setPriority(Request<?> request, UrlRequest.Builder builder) {
+        builder.setPriority(request.getPriority().ordinal());
+    }
+
+    private int getContentLength(UrlResponseInfo urlResponseInfo) {
```
jpd236, line 276: So looking more closely at how BasicNetwork#inputStreamToBytes works, I still don't quite think we've matched it.

What that method does is grab a 1K buffer from the pool and use it to read chunks of the InputStream into a PoolingByteArrayOutputStream created with the Content-Length as the expected size. What we're doing here is creating a direct buffer with the Content-Length (or 1K if unknown) to read chunks into an unsized PoolingByteArrayOutputStream (which defaults to 256 bytes).

It sounds like we can't use the pool for the ByteBuffer we use to read chunks of data from Cronet, but I think we should size it at 1K to match BasicNetwork. The response size should come into play for the PoolingByteArrayOutputStream we use to accumulate chunks of responses, rather than relying on the default size for that.

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -0,0 +1,113 @@
...
+        } else {
+            throw entry.get().authFailureError;
+        }
+    }
+
+    private static class Response {
+        HttpResponse httpResponse;
+        IOException ioException;
+        AuthFailureError authFailureError;
+        int type;
```
jpd236, line 100: You could use an @IntDef for this to get better type safety without the cost of an enum (which would normally be preferred), but then I'm not sure why you need this field at all. Can't you just do:

```java
Response response = entry.get();
if (response.httpResponse != null) {
  return response.httpResponse;
} else if (response.ioException != null) {
  ...
}
...
```

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -0,0 +1,113 @@
...
+    public abstract void executeRequest(
+            Request<?> request, Map<String, String> additionalHeaders, OnRequestComplete callback);
+
+    /**
+     * Performs an HTTP request with the given parameters.
+     *
+     * @param request the request to perform
+     * @param additionalHeaders additional headers to be sent together with {@link
+     *     Request#getHeaders()}
+     * @return the {@link HttpResponse}, or null if an InterruptedException occurs.
```
jpd236, line 43: I don't think callers expect null responses, do they? We can't break the base interface we're implementing here by doing something undocumented/unexpected there. InterruptedIOException might be a reasonable choice here.

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
sphill99: Yeah, that works. I thought having a type parameter would make things simpler, but I think it obscures it.

jpd236: we do something similar in RequestFuture

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
sphill99: Okay, that should be a better option.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: hmm that's interesting. I printed all the headers and they were all lower case. And using "content-length" does seem to be working right now. I'll do some more testing with it.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: I see what you mean, but we already have the HTTP response by the time inputStreamToBytes is called. Only thing I can think of is setting the byteArrayOutputStream once we have read the headers, but there's an issue there since that's within an inner class. I used an array to get around the variable needing to be final, but I think it's a bit weird.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: I ended up adding a Builder class, and allowing the caller to set their own CronetEngine, but defaulting to CronetEngine.build(context).

sphill99:
I believe all comments have now been addressed.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,300 @@
...
+import org.chromium.net.UrlRequest;
+import org.chromium.net.UrlRequest.Callback;
+import org.chromium.net.UrlResponseInfo;
+
+/**
+ * A {@link AsyncHttpStack} that's based on Cronet's fully asynchronous API for network requests.
+ */
+public class CronetHttpStack extends AsyncHttpStack {
+
+    private static final int DEFAULT_POOL_SIZE = 4096;
```
sophiahaim, line 37: nit: I think this can be removed; it doesn't look like it's used (since you're setting the size in Builder now)

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -0,0 +1,110 @@
...
+            throws IOException, AuthFailureError {
+        final CountDownLatch latch = new CountDownLatch(1);
+        final AtomicReference<Response> entry = new AtomicReference<>();
+        executeRequest(
+                request,
+                additionalHeaders,
+                new OnRequestComplete() {
+                    @Override
+                    public void onSuccess(HttpResponse httpResponse) {
+                        Response response = new Response(httpResponse, null, null);
```
sophiahaim, line 60: nit: Add param comments since it's not clear what the nulls are:
new Response(httpResponse, /* ioException= */ null, /* authFailureError= */ null)

same for the calls below

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: Oh yeah, that was moved into Builder

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
+package com.android.volley.cronet;
```
jpd236, line 1: just realized I forgot about this for previous new files, but all Java files should have a copyright header at the top (matching existing ones, with the year updated).

for this PR, just handle the new files in this class, but could you also send a follow-up adding it for the new files you've added in previous PRs?

src/main/java/com/android/volley/cronet/CronetHttpStack.java
jpd236: Historically the priority has only been used to determine the order in which requests get pulled from Volley's internal queue and not propagated to the underlying HTTP library itself. That said, this is probably more because it wasn't really possible/meaningful before, since we blocked network threads on completing requests rather than sending a bunch of requests of different priorities to the queue and waiting for them to get back.

So I think it's actually probably reasonable to push the Request priority down to Cronet in addition to using it to prioritize requests in Volley's own queues. If need be, we could always provide a hook to clients to turn that off (or they could leave priorities at normal to make them all equivalent), but I'd wait until we hear a concrete use case for that.

Feel free to do this in a small follow up if it's non-trivial to convert between the priorities to prevent this PR from dragging on.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+ */
+public class CronetHttpStack extends AsyncHttpStack {
+
+    private final CronetEngine mCronetEngine;
+    private final Executor mCallbackExecutor;
+    private final ExecutorService mBlockingExecutor;
+    private final ByteArrayPool mPool;
+    private final UrlRewriter mUrlRewriter;
+
+    private CronetHttpStack(Builder builder) {
```
jpd236, line 43: nit: My recommendation would be to have the constructor take the arguments directly that it expects, and have the Builder call the constructor with those arguments.

This is the more conventional pattern that I've seen for this sort of thing, and the benefit of that approach is that it makes unit testing the class simpler as you don't need to use the builder interface just to create a test instance.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+            final Map<String, String> additionalHeaders,
+            final OnRequestComplete callback) {
+        final Callback urlCallback =
+                new Callback() {
+                    PoolingByteArrayOutputStream bytesReceived = null;
+                    WritableByteChannel receiveChannel = null;
+
+                    @Override
+                    public void onRedirectReceived(
+                            UrlRequest urlRequest, UrlResponseInfo urlResponseInfo, String s) {
```
jpd236, line 63: nit:  what is "s"? Assuming it has a more descriptive name in the Cronet documentation, could we use that name for clarity?

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+                            UrlResponseInfo urlResponseInfo,
+                            CronetException e) {
+                        callback.onError(e);
+                    }
+                };
+
+        String url = request.getUrl();
+        String rewritten = mUrlRewriter.rewriteUrl(url);
+        if (rewritten == null) {
+            callback.onError(new IOException("URL blocked by rewriter: " + url));
```
jpd236, line 117: you need to return here to prevent the rest of the request from being attempted

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+        url = rewritten;
+
+        // We can call allowDirectExecutor here and run directly on the network thread, since all
+        // the callbacks are non-blocking.
+        final UrlRequest.Builder builder =
+                mCronetEngine
+                        .newUrlRequestBuilder(url, urlCallback, mCallbackExecutor)
+                        .allowDirectExecutor()
+                        .disableCache();
+        // This code may be blocking, so submit it to the blocking executor.
```
jpd236, line 128: nit: replace "This code" with "request.getHeaders()" to be more explicit.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+        private static final int DEFAULT_POOL_SIZE = 4096;
+        private CronetEngine mCronetEngine;
+        private Executor mCallbackExecutor;
+        private ExecutorService mBlockingExecutor;
+        private ByteArrayPool mPool;
+        private UrlRewriter mUrlRewriter;
+
+        public Builder(
+                Context context, ExecutorService blockingExecutor, Executor callbackExecutor) {
+            mCronetEngine = new CronetEngine.Builder(context).build();
```
jpd236, line 255: My recommendation would be to defer this work to build() for any field not explicitly set. Otherwise, we're creating unnecessary CronetEngine, ByteArrayPool, and UrlRewriter objects that just get discarded.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+                    new UrlRewriter() {
+                        @Override
+                        public String rewriteUrl(String originalUrl) {
+                            return originalUrl;
+                        }
+                    };
+        }
+
+        /**
+         * Sets the CronetEngine to be used. Would recommend disabling the HTTP cache, as it is
```
jpd236, line 269: does this recommendation still hold now that we call disableCache() on a per-request basis above?

(which I actually think is a better approach, now that I see it, since it allows an app to share a CronetEngine that might be used outside of Volley where the cache layer *is* desirable).

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+         * redundant with Volley's.
+         */
+        public Builder setCronetEngine(CronetEngine engine) {
+            mCronetEngine = engine;
+            return this;
+        }
+
+        /** Sets the ByteArrayPool to be used. */
```
jpd236, line 277: document the default behavior if this is uncalled (we use a new pool that starts at 4K)

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+     */
+    public static class Builder {
+        private static final int DEFAULT_POOL_SIZE = 4096;
+        private CronetEngine mCronetEngine;
+        private Executor mCallbackExecutor;
+        private ExecutorService mBlockingExecutor;
+        private ByteArrayPool mPool;
+        private UrlRewriter mUrlRewriter;
+
+        public Builder(
```
jpd236, line 253: I'd recommend setting the blockingExecutor/callbackExecutor in their own set methods as well.

Even if they're required, think about how likely it is for a caller to accidentally do:

ExecutorService blockingExecutor = ...
ExecutorService callbackExecutor = ...
new Builder(context, callbackExecutor, blockingExecutor).build();

vs.

new Builder(context).setCallbackExecutor(blockingExecutor).setBlockingExecutor(callbackExecutor).build();

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+                Context context, ExecutorService blockingExecutor, Executor callbackExecutor) {
+            mCronetEngine = new CronetEngine.Builder(context).build();
+            mCallbackExecutor = callbackExecutor;
```
jpd236, line 256: you should validate that these executors are non-null

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,293 @@
...
+    public static class Builder {
+        private static final int DEFAULT_POOL_SIZE = 4096;
+        private CronetEngine mCronetEngine;
+        private Executor mCallbackExecutor;
+        private ExecutorService mBlockingExecutor;
+        private ByteArrayPool mPool;
+        private UrlRewriter mUrlRewriter;
+
+        public Builder(
+                Context context, ExecutorService blockingExecutor, Executor callbackExecutor) {
```
jpd236, line 254: I feel like we should just require an ExecutorService for the callback; it's extremely rare to have an Executor but not an ExecutorService, and it would give us more flexibility if we need it.

src/main/java/com/android/volley/toolbox/UrlRewriter.java
```diff
@@ -0,0 +1,10 @@
+package com.android.volley.toolbox;
+
+/** An interface for transforming URLs before use. */
+public interface UrlRewriter {
+    /**
+     * Returns a URL to use instead of the provided one, or null to indicate this URL should not be
+     * used at all.
+     */
+    String rewriteUrl(String originalUrl);
```
jpd236, line 9: add `@Nullable`

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: oh yeah, I'll add those to all the new files.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: Okay, that makes sense. I saw two different ways to do it, but I'll change it.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: What's should be done if they are? Throw an illegalArgumentException?

jpd236: yep

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: Yeah, I don't think it makes sense since we call disableCache().

sphill99:
Every comment should have been addressed. I did add the setPriority for the UrlRequest, since there were only 4 priorities for a Volley request, and 5 options for the UrlRequest.Builder. I figured the priority should stay the same if we pass that along, but not sure if there could be an issue since their defaults are different. I also reworked the Builder class a good amount, based on the comments above.

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -0,0 +1,338 @@
...
+    /** Sets the UploadDataProvider of the UrlRequest.Builder */
+    private void addBodyIfExists(@Nullable byte[] body, UrlRequest.Builder builder) {
+        if (body != null) {
+            UploadDataProvider dataProvider = UploadDataProviders.create(body);
+            builder.setUploadDataProvider(dataProvider, mCallbackExecutor);
+        }
+    }
+
+    private void setRequestPriority(Request<?> request, UrlRequest.Builder builder) {
+        builder.setPriority(request.getPriority().ordinal() + 1);
```
jpd236, line 272: this code strikes me as risky; nothing guarantees that we wouldn't add a new priority value or reorder them in some form, unlikely though it may be.

it's also not obvious to me that we should map in this manner, just because the order is relatively the same. In particular, we map "NORMAL" in Volley to "LOW" in Cronet, which is counterintuitive, and if you're sharing the Cronet queue with stuff outside Volley, would result in all of Volley's seemingly normal requests being deprioritized. I think it would be more intuitive to map NORMAL -> DEFAULT and combine HIGH/IMMEDIATE into REQUEST_PRIORITY_HIGHEST since Cronet doesn't offer two levels above DEFAULT.

I think it would be better to define a helper method like:

```java
private int getPriority(Request<?> request) {
  switch (request.getPriority()) {
    case LOW:
      return REQUEST_PRIORITY_LOW;
    case HIGH:
    case IMMEDIATE:
      return REQUEST_PRIORITY_HIGHEST;
    case MEDIUM:
    default:
      return REQUEST_PRIORITY_NORMAL;
    }
}
```

and call builder.setPriority(getPriority(request)); inline instead.

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @param request to perform
+     * @param additionalHeaders to be sent together with {@link Request#getHeaders()}
+     * @param callback to be called after retrieving the {@link HttpResponse} or throwing an error.
+     */
+    public abstract void executeRequest(
+            Request<?> request, Map<String, String> additionalHeaders, OnRequestComplete callback);
+
+    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
+    public abstract void setCallbackExecutor(ExecutorService executorService);
```
jpd236, line 57: could you add Javadocs for these methods? Anyone implementing a subclass needs to know how to implement these (e.g. that they'll be called before being asked to execute any requests, and to note how the executors should be used).

Also, I think "setNonBlockingExecutor" would be more consistent with what we were planning to call the executor elsewhere, and more flexible (it's not necessarily tied to callbacks - the important thing is that it doesn't run blocking work).

src/main/java/com/android/volley/cronet/CronetHttpStack.java
sphill99: this is definitely cleaner and safer. I'll make that change.

sphill99:
I made those two changes, as well as switching the name of the method to "setNonBlockingExecutor."

PR #357, opened 2020-08-11
AsyncNetwork implementation as well as unit testing

sphill99:
This PR defines the AsyncNetwork class, as well as an implementation in the form of BasicAsyncNetwork. All test cases that were covered for BasicNetwork are covered here as unit tests. Some shared code I put into a NetworkUtility class.

Some open questions/comments:
- BasicAsyncNetworkTest and BasicNetwork share one helper method. Wasn't sure if it was worth adding a shared class to not duplicate code.
- Wasn't totally sure if the handling of timeouts before retry is what we're looking for.
- For this to compile, I had to change AsyncHttpStack's setExecutor methods to require(LIBRARY) instead of also including subgroups. I'm looking into this now, but changed it for the time being. Also, since the RequestQueue doesn't see the stack directly, I have methods in the network set the executors in AsyncHttpStack.

### DIFF
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -32,7 +32,7 @@ public abstract class AsyncNetwork implements Network {
         mAsyncStack = stack;
     }
 
-    /** Inteface for callback to be called after request is processed */
+    /** Inteface for callback to be called after request is processed. */
     public interface OnRequestComplete {
         /** Method to be called after successful network request. */
         void onSuccess(NetworkResponse networkResponse);
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -16,13 +16,21 @@
 
 package com.android.volley;
 
-import androidx.annotation.Nullable;
 import androidx.annotation.RestrictTo;
+import com.android.volley.toolbox.AsyncHttpStack;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicReference;
 
+/** An asynchronous implementation of {@link Network} to perform requests. */
 public abstract class AsyncNetwork implements Network {
+    protected AsyncHttpStack mAsyncStack;
+    protected ExecutorService mBlockingExecutor;
+    protected ExecutorService mNonBlockingExecutor;
+
+    protected AsyncNetwork(AsyncHttpStack stack) {
+        mAsyncStack = stack;
+    }
 
     public interface OnRequestComplete {
         void onSuccess(NetworkResponse networkResponse);
@@ -30,25 +38,39 @@ public abstract class AsyncNetwork implements Network {
         void onError(VolleyError volleyError);
     }
 
+    /**
+     * Non-blocking method to perform the specified request.
+     *
+     * @param request Request to process
+     * @param callback to be called once NetworkResponse is received
+     */
     public abstract void performRequest(Request<?> request, OnRequestComplete callback);
 
+    /**
+     * Blocking method to perform network request.
+     *
+     * @param request Request to process
+     * @return response retrieved from the network
+     * @throws VolleyError in the event of an error
+     */
     @Override
     public NetworkResponse performRequest(Request<?> request) throws VolleyError {
         // Is there a better way to go about this?
         final CountDownLatch latch = new CountDownLatch(1);
-        final AtomicReference<Response> entry = new AtomicReference<>();
+        final AtomicReference<NetworkResponse> response = new AtomicReference<>();
+        final AtomicReference<VolleyError> error = new AtomicReference<>();
         performRequest(
                 request,
                 new OnRequestComplete() {
                     @Override
                     public void onSuccess(NetworkResponse networkResponse) {
-                        entry.set(new Response(networkResponse, /* error= */ null));
+                        response.set(networkResponse);
                         latch.countDown();
                     }
 
                     @Override
                     public void onError(VolleyError volleyError) {
-                        entry.set(new Response(/* networkResponse= */ null, volleyError));
+                        error.set(volleyError);
                         latch.countDown();
                     }
                 });
@@ -60,11 +82,10 @@ public abstract class AsyncNetwork implements Network {
             throw new VolleyError(e);
         }
 
-        Response response = entry.get();
-        if (response.networkResponse != null) {
-            return response.networkResponse;
-        } else if (response.volleyError != null) {
-            throw response.volleyError;
+        if (response.get() != null) {
+            return response.get();
+        } else if (error.get() != null) {
+            throw error.get();
         } else {
             throw new VolleyError("Neither response entry was set");
         }
@@ -77,7 +98,10 @@ public abstract class AsyncNetwork implements Network {
      * AsyncHttpStack.
      */
     @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
-    public abstract void setNonBlockingExecutorForStack(ExecutorService executor);
+    public void setNonBlockingExecutorForStack(ExecutorService executor) {
+        mNonBlockingExecutor = executor;
+        mAsyncStack.setNonBlockingExecutor(executor);
+    }
 
     /**
      * This method sets the blocking executor to be used by the network and stack for potentially
@@ -86,15 +110,8 @@ public abstract class AsyncNetwork implements Network {
      * com.android.volley.toolbox.AsyncHttpStack}
      */
     @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
-    public abstract void setBlockingExecutor(ExecutorService executor);
-
-    static class Response {
-        @Nullable NetworkResponse networkResponse;
-        @Nullable VolleyError volleyError;
-
-        private Response(@Nullable NetworkResponse networkResponse, @Nullable VolleyError error) {
-            this.networkResponse = networkResponse;
-            volleyError = error;
-        }
+    public void setBlockingExecutor(ExecutorService executor) {
+        mBlockingExecutor = executor;
+        mAsyncStack.setBlockingExecutor(executor);
     }
 }
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -25,8 +25,8 @@ import java.util.concurrent.atomic.AtomicReference;
 /** An asynchronous implementation of {@link Network} to perform requests. */
 public abstract class AsyncNetwork implements Network {
     protected AsyncHttpStack mAsyncStack;
-    protected ExecutorService mBlockingExecutor;
-    protected ExecutorService mNonBlockingExecutor;
+    private ExecutorService mBlockingExecutor;
+    private ExecutorService mNonBlockingExecutor;
 
     protected AsyncNetwork(AsyncHttpStack stack) {
         mAsyncStack = stack;
@@ -113,5 +113,13 @@ public abstract class AsyncNetwork implements Network {
     public void setBlockingExecutor(ExecutorService executor) {
         mBlockingExecutor = executor;
         mAsyncStack.setBlockingExecutor(executor);
+    }
+
+    protected ExecutorService getBlockingExecutor() {
+        return mBlockingExecutor;
+    }
+
+    protected ExecutorService getNonBlockingExecutor() {
+        return mNonBlockingExecutor;
     }
 }
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -32,9 +32,12 @@ public abstract class AsyncNetwork implements Network {
         mAsyncStack = stack;
     }
 
+    /** Inteface for callback to be called after request is processed */
     public interface OnRequestComplete {
+        /** Method to be called after successful network request. */
         void onSuccess(NetworkResponse networkResponse);
 
+        /** Method to be called after unsuccessful network request. */
         void onError(VolleyError volleyError);
     }
 
@@ -91,11 +94,11 @@ public abstract class AsyncNetwork implements Network {
     }
 
     /**
-     * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
-     * This method must be called before performing any requests.
+     * This method sets the non blocking executor to be used by the network and stack for
+     * non-blocking tasks. This method must be called before performing any requests.
      */
     @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
-    public void setNonBlockingExecutorForStack(ExecutorService executor) {
+    public void setNonBlockingExecutor(ExecutorService executor) {
         mNonBlockingExecutor = executor;
         mAsyncStack.setNonBlockingExecutor(executor);
     }
@@ -110,14 +113,17 @@ public abstract class AsyncNetwork implements Network {
         mAsyncStack.setBlockingExecutor(executor);
     }
 
+    /** Gets blocking executor to perform any potentially blocking tasks. */
     protected ExecutorService getBlockingExecutor() {
         return mBlockingExecutor;
     }
 
+    /** Gets non-blocking executor to perform any non-blocking tasks. */
     protected ExecutorService getNonBlockingExecutor() {
         return mNonBlockingExecutor;
     }
 
+    /** Gets the {@link AsyncHttpStack} to be used by the network. */
     protected AsyncHttpStack getHttpStack() {
         return mAsyncStack;
     }
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -63,8 +63,10 @@ public abstract class AsyncNetwork implements Network {
         Response response = entry.get();
         if (response.networkResponse != null) {
             return response.networkResponse;
-        } else {
+        } else if (response.volleyError != null) {
             throw response.volleyError;
+        } else {
+            throw new VolleyError("Neither response entry was set");
         }
     }
 
@@ -74,7 +76,7 @@ public abstract class AsyncNetwork implements Network {
      * nothing. This method must be called before performing any requests if you are using an
      * AsyncHttpStack.
      */
-    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
     public abstract void setNonBlockingExecutorForStack(ExecutorService executor);
 
     /**
@@ -83,12 +85,12 @@ public abstract class AsyncNetwork implements Network {
      * executor for the stack if it is an instance of {@link
      * com.android.volley.toolbox.AsyncHttpStack}
      */
-    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
     public abstract void setBlockingExecutor(ExecutorService executor);
 
     static class Response {
-        NetworkResponse networkResponse;
-        VolleyError volleyError;
+        @Nullable NetworkResponse networkResponse;
+        @Nullable VolleyError volleyError;
 
         private Response(@Nullable NetworkResponse networkResponse, @Nullable VolleyError error) {
             this.networkResponse = networkResponse;
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -24,7 +24,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 /** An asynchronous implementation of {@link Network} to perform requests. */
 public abstract class AsyncNetwork implements Network {
-    protected AsyncHttpStack mAsyncStack;
+    private final AsyncHttpStack mAsyncStack;
     private ExecutorService mBlockingExecutor;
     private ExecutorService mNonBlockingExecutor;
 
@@ -55,7 +55,6 @@ public abstract class AsyncNetwork implements Network {
      */
     @Override
     public NetworkResponse performRequest(Request<?> request) throws VolleyError {
-        // Is there a better way to go about this?
         final CountDownLatch latch = new CountDownLatch(1);
         final AtomicReference<NetworkResponse> response = new AtomicReference<>();
         final AtomicReference<VolleyError> error = new AtomicReference<>();
@@ -92,10 +91,8 @@ public abstract class AsyncNetwork implements Network {
     }
 
     /**
-     * This method sets the non blocking executor to be used by the stack for non-blocking tasks. If
-     * you are not using an {@link com.android.volley.toolbox.AsyncHttpStack}, this should do
-     * nothing. This method must be called before performing any requests if you are using an
-     * AsyncHttpStack.
+     * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
+     * This method must be called before performing any requests if you are using an AsyncHttpStack.
      */
     @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
     public void setNonBlockingExecutorForStack(ExecutorService executor) {
@@ -105,9 +102,7 @@ public abstract class AsyncNetwork implements Network {
 
     /**
      * This method sets the blocking executor to be used by the network and stack for potentially
-     * blocking tasks. This method must be called before performing any requests. Only set the
-     * executor for the stack if it is an instance of {@link
-     * com.android.volley.toolbox.AsyncHttpStack}
+     * blocking tasks. This method must be called before performing any requests.
      */
     @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
     public void setBlockingExecutor(ExecutorService executor) {
@@ -121,5 +116,9 @@ public abstract class AsyncNetwork implements Network {
 
     protected ExecutorService getNonBlockingExecutor() {
         return mNonBlockingExecutor;
+    }
+
+    protected AsyncHttpStack getHttpStack() {
+        return mAsyncStack;
     }
 }
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.android.volley;
+
+import androidx.annotation.Nullable;
+import androidx.annotation.RestrictTo;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AsyncNetwork implements Network {
+
+    public interface OnRequestComplete {
+        void onSuccess(NetworkResponse networkResponse);
+
+        void onError(VolleyError volleyError);
+    }
+
+    public abstract void performRequest(Request<?> request, OnRequestComplete callback);
+
+    @Override
+    public NetworkResponse performRequest(Request<?> request) throws VolleyError {
+        // Is there a better way to go about this?
+        final CountDownLatch latch = new CountDownLatch(1);
+        final AtomicReference<Response> entry = new AtomicReference<>();
+        performRequest(
+                request,
+                new OnRequestComplete() {
+                    @Override
+                    public void onSuccess(NetworkResponse networkResponse) {
+                        entry.set(new Response(networkResponse, /* error= */ null));
+                        latch.countDown();
+                    }
+
+                    @Override
+                    public void onError(VolleyError volleyError) {
+                        entry.set(new Response(/* networkResponse= */ null, volleyError));
+                        latch.countDown();
+                    }
+                });
+        try {
+            latch.await();
+        } catch (InterruptedException e) {
+            VolleyLog.e(e, "while waiting for CountDownLatch");
+            Thread.currentThread().interrupt();
+            throw new VolleyError(e);
+        }
+
+        Response response = entry.get();
+        if (response.networkResponse != null) {
+            return response.networkResponse;
+        } else {
+            throw response.volleyError;
+        }
+    }
+
+    /**
+     * This method sets the non blocking executor to be used by the stack for non-blocking tasks. If
+     * you are not using an {@link com.android.volley.toolbox.AsyncHttpStack}, this should do
+     * nothing. This method must be called before performing any requests if you are using an
+     * AsyncHttpStack.
+     */
+    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
+    public abstract void setNonBlockingExecutorForStack(ExecutorService executor);
+
+    /**
+     * This method sets the blocking executor to be used by the network and stack for potentially
+     * blocking tasks. This method must be called before performing any requests. Only set the
+     * executor for the stack if it is an instance of {@link
+     * com.android.volley.toolbox.AsyncHttpStack}
+     */
+    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
+    public abstract void setBlockingExecutor(ExecutorService executor);
+
+    static class Response {
+        NetworkResponse networkResponse;
+        VolleyError volleyError;
+
+        private Response(@Nullable NetworkResponse networkResponse, @Nullable VolleyError error) {
+            this.networkResponse = networkResponse;
+            volleyError = error;
+        }
+    }
+}
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -32,7 +32,7 @@ public abstract class AsyncNetwork implements Network {
         mAsyncStack = stack;
     }
 
-    /** Inteface for callback to be called after request is processed. */
+    /** Interface for callback to be called after request is processed. */
     public interface OnRequestComplete {
         /** Method to be called after successful network request. */
         void onSuccess(NetworkResponse networkResponse);
--- a/src/main/java/com/android/volley/AsyncNetwork.java
+++ b/src/main/java/com/android/volley/AsyncNetwork.java
@@ -92,7 +92,7 @@ public abstract class AsyncNetwork implements Network {
 
     /**
      * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
-     * This method must be called before performing any requests if you are using an AsyncHttpStack.
+     * This method must be called before performing any requests.
      */
     @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
     public void setNonBlockingExecutorForStack(ExecutorService executor) {
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -65,7 +65,7 @@ public class CronetHttpStack extends AsyncHttpStack {
             final Map<String, String> additionalHeaders,
             final OnRequestComplete callback) {
         if (getBlockingExecutor() == null || getNonBlockingExecutor() == null) {
-            throw new IllegalStateException("Must set blocking / non-blocking executors");
+            throw new IllegalStateException("Must set blocking and non-blocking executors");
         }
         final Callback urlCallback =
                 new Callback() {
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -17,8 +17,9 @@
 package com.android.volley.cronet;
 
 import android.content.Context;
+
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import androidx.annotation.RestrictTo;
 import androidx.annotation.VisibleForTesting;
 import com.android.volley.AuthFailureError;
 import com.android.volley.Header;
@@ -36,7 +37,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
 import org.chromium.net.CronetEngine;
 import org.chromium.net.CronetException;
 import org.chromium.net.UploadDataProvider;
@@ -51,16 +51,12 @@ import org.chromium.net.UrlResponseInfo;
 public class CronetHttpStack extends AsyncHttpStack {
 
     private final CronetEngine mCronetEngine;
-    private ExecutorService mCallbackExecutor;
-    private ExecutorService mBlockingExecutor;
     private final ByteArrayPool mPool;
     private final UrlRewriter mUrlRewriter;
 
     private CronetHttpStack(
             CronetEngine cronetEngine, ByteArrayPool pool, UrlRewriter urlRewriter) {
         mCronetEngine = cronetEngine;
-        mCallbackExecutor = null;
-        mBlockingExecutor = null;
         mPool = pool;
         mUrlRewriter = urlRewriter;
     }
@@ -70,7 +66,7 @@ public class CronetHttpStack extends AsyncHttpStack {
             final Request<?> request,
             final Map<String, String> additionalHeaders,
             final OnRequestComplete callback) {
-        if (mBlockingExecutor == null || mCallbackExecutor == null) {
+        if (getBlockingExecutor() == null || getNonBlockingExecutor() == null) {
             throw new IllegalStateException("Must set blocking / non-blocking executors");
         }
         final Callback urlCallback =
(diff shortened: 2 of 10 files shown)


src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -136,7 +126,8 @@ public NetworkResponse performRequest(Request<?> request) throws VolleyError {
                                 responseHeaders);
```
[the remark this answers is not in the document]
sphill99: I had not considered that, but I'll look into it.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,419 @@
...
+        try {
+            retryPolicy.retry(exception);
+        } catch (VolleyError e) {
+            request.addMarker(
+                    String.format("%s-timeout-giveup [timeout=%s]", logPrefix, oldTimeout));
+            callback.onError(e);
+            return;
+        }
+        request.addMarker(String.format("%s-retry [timeout=%s]", logPrefix, oldTimeout));
+        // If we are using an async stack, then perform retry after a timeout.
```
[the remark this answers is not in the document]
sphill99: I had also considered having the network user set the timeout for retrying again. Then we could also check which stack they're using there instead of in this method, and just default to 0 if non-async stack. Wasn't sure though about this.

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,100 @@
...
+
+package com.android.volley;
+
+import androidx.annotation.Nullable;
+import androidx.annotation.RestrictTo;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class AsyncNetwork implements Network {
```
jpd236, line 25: Need Javadocs for any public APIs

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,100 @@
...
+        void onSuccess(NetworkResponse networkResponse);
+
+        void onError(VolleyError volleyError);
+    }
+
+    public abstract void performRequest(Request<?> request, OnRequestComplete callback);
+
+    @Override
+    public NetworkResponse performRequest(Request<?> request) throws VolleyError {
+        // Is there a better way to go about this?
```
jpd236, line 37: looks mostly fine to me, but personally I'd probably scrap the Response POJO and just have two AtomicReferences, one for NetworkResponse and one for VolleyError. It doesn't seem like the container is buying you much.

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -57,14 +57,14 @@ public abstract void executeRequest(
      * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
      * This method must be called before executing any requests.
      */
-    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
```
jpd236, line 60: The problem here is that it prevents someone outside the Volley from extending AsyncHttpStack or AsyncNetwork, which is totally valid. Sounds like you're still looking into why this is happening, but let me know if I can help.

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * This method sets the non blocking executor to be used by the stack for non-blocking tasks. If
+     * you are not using an {@link com.android.volley.toolbox.AsyncHttpStack}, this should do
+     * nothing. This method must be called before performing any requests if you are using an
+     * AsyncHttpStack.
+     */
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
+    public abstract void setNonBlockingExecutorForStack(ExecutorService executor);
```
jpd236, line 80: WDYT about implementing these methods in AsyncNetwork directly rather than requiring subclasses to do it? E.g. having a protected constructor here which takes the AsyncHttpStack and does the right thing? I'm not sure why any subclass would do anything differently.

You can make the executors accessible to subclasses either by making the variables protected here or (preferably) exposing protected getters.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+
+/**
+ * Utility class for methods that are shared between {@link BasicNetwork} and {@link
+ * BasicAsyncNetwork}
+ */
+public class NetworkUtility {
+    protected static final boolean DEBUG = VolleyLog.DEBUG;
```
jpd236, line 37: This seems a bit redundant, vs. just referencing VolleyLog.DEBUG directly.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private static final int SLOW_REQUEST_THRESHOLD_MS = 3000;
+    /**
+     * Converts Headers[] to Map&lt;String, String&gt;.
+     *
+     * @deprecated Should never have been exposed in the API. This method may be removed in a future
+     *     release of Volley.
+     */
+    @Deprecated
+    static Map<String, String> convertHeaders(Header[] headers) {
```
jpd236, line 47: This has to remain where it was - per the Javadoc, it was inadvertently exposed as a publicly accessible API, so moving it might break users of the library.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * <p>An HTTP 304 response does not have all header fields. We have to use the header fields
+     * from the cache entry plus the new ones from the response. See also:
+     * http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5
+     *
+     * @param responseHeaders Headers from the network response.
+     * @param entry The cached response.
+     * @return The combined list of headers.
+     */
+    static List<Header> combineHeaders(List<Header> responseHeaders, Cache.Entry entry) {
```
jpd236, line 66: Some of these header utilities might be good fits for https://github.com/google/volley/blob/master/src/main/java/com/android/volley/toolbox/HttpHeaderParser.java, rather than needing a new class

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
```diff
@@ -0,0 +1,431 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
+public class BasicAsyncNetworkTest {
+
+    @Mock private Request<String> mMockRequest;
+    @Mock private RetryPolicy mMockRetryPolicy;
+    private ExecutorService executor = Executors.newSingleThreadExecutor();
```
jpd236, line 68: Generally speaking, we want to avoid using real threads in unit tests as we don't have as much control over when they run. See http://robolectric.org/javadoc/4.3/org/robolectric/android/util/concurrent/InlineExecutorService.html, for example, or just use MoreExecutors.directExecutor() if that works?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+        if (mBaseHttpStack instanceof AsyncHttpStack) {
+            mHandler.postDelayed(
```
jpd236, line 388: We shouldn't need a Handler for this; we should be able to use the Executor. But we might need to change it to a ScheduledExecutorService instead if we decide to keep this behavior.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+
+    protected final Handler mHandler;
+
+    /**
+     * @param httpStack HTTP stack to be used
+     * @deprecated use {@link #BasicAsyncNetwork(BaseHttpStack)} instead to avoid depending on
+     *     Apache HTTP. This method may be removed in a future release of Volley.
+     */
+    @Deprecated
+    public BasicAsyncNetwork(HttpStack httpStack) {
```
jpd236, line 73: let's use the Builder pattern here too?

We don't need all these deprecated constructors for a new class. Just don't accept HttpStack from BaseAsyncNetwork.

For that matter, do we actually need to support legacy BaseHttpStacks from BasicAsyncNetwork? Or can we just require that (Basic)Network be used instead in that case?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+                responseContents = new byte[0];
+            } else {
+                Runnable run =
+                        new Runnable() {
+                            @Override
+                            public void run() {
+                                PoolingByteArrayOutputStream bytes =
+                                        new PoolingByteArrayOutputStream(
+                                                mPool, httpResponse.getContentLength());
+                                byte[] buffer = new byte[1024];
```
jpd236, line 163: Shouldn't this come from the pool?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+        byte[] responseContents = httpResponse.getContentBytes();
+
+        if (responseContents == null) {
+            final InputStream inputStream = httpResponse.getContent();
+            if (inputStream == null) {
+                // Add 0 byte response as a way of honestly representing a
+                // no-content request.
+                responseContents = new byte[0];
+            } else {
+                Runnable run =
```
jpd236, line 156: A lot of this is copied from the BasicNetwork stream copying code, right? Can we share it?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+                                        callback,
+                                        responseHeaders,
+                                        finalResponseContents);
+                            }
+                        };
+                mBlockingExecutor.execute(run);
+                return;
+            }
+        }
+        runAfterBytesReceived(
```
jpd236, line 194: nit: since this only applies to the case that the response bytes are already read, which is checked many lines above, I'd put this up there and short-circuit return in that case, and then do the rest of the handling for the stream reading case.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+                                                callback,
+                                                e,
+                                                requestStartMs,
+                                                httpResponse,
+                                                bytes.toByteArray());
+                                    }
+                                    bytes.write(buffer, 0, count);
+                                }
+                                byte[] finalResponseContents = bytes.toByteArray();
+                                runAfterBytesReceived(
```
jpd236, line 180: nit: perhaps "onResponseRead", which would be consistent with "onRequestSucceeded" or "onRequestFailed" naming-wise.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,432 @@
...
+                statusCode,
+                httpResponse,
+                request,
+                callback,
+                responseHeaders,
+                responseContents);
+    }
+
+    /* Method to be called after a failed network request */
+    private void onRequestFailed(
```
jpd236, line 205: any more code sharing we can do here w/ BasicNetwork to go from Exception type to how it gets handled?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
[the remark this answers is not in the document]
jpd236: It looks like what the current code is doing is just trying again after an unconfigurable 100ms delay, which doesn't seem right to me.

What I think we should do for now is match the existing behavior as much as possible with our current APIs. That would mean retrying immediately upon failure. In theory, when the request is retried, getTimeoutMs will return an increased time unit when passed to Cronet. However, as we established earlier, Cronet doesn't actually support setting a timeout, so this increased timeout gets ignored. But that's an implementation detail of the HTTP stack, and other stacks might respect the timeout.

I do think we should make it possible for clients to configure a gap between retries, as the doc noted. But this would be a separate concept which requires a new API, and I don't think we should assume a fixed 100ms retry interval for each retry.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: If what we're passing in is a ThreadPoolExecutor, how could we change it to a ScheduledExecutorService?

src/main/java/com/android/volley/toolbox/NetworkUtility.java
sphill99: Yeah, I agree. I had just kept that from BasicNetwork, but doesn't really seem to have a purpose.

src/main/java/com/android/volley/AsyncNetwork.java
sphill99: Yeah, I like this a lot. I can't another implementation for this method.

src/main/java/com/android/volley/AsyncNetwork.java
sphill99: Okay, that's what I had, but switched it to keep the same structure as AsyncHttpStack.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: Yep

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: Are you saying our constructor should only take in an AsyncHttpStack? My only worry is then we have to handle both AsyncNetworks and Networks in AsyncRequestQueue. Unless we still require users to pass in an AsyncNetwork into AsyncRequestQueue. Honestly, if you're using a blocking network and blocking stack (and potentially a blocking cache), does it make sense for them to be able to use the AsyncRequestQueue? With the cache is makes sense, since we require a newish api level, but in this case could we just require they use an AsyncNetwork, which requires an AsyncStack?

jpd236: Yes, I agree with what you're saying - I don't see any reason to support the blocking flows for Network in the Async stack. Cache is indeed needed for older devices as you say, but we have an AsyncNetwork implementation that should be usable by anyone willing to use AsyncRequestQueue (or if you don't like Cronet, you could use OkHttp instead, which would also work fine), and if you're using a Network, that broadly defeats the purpose of using the async stack to begin with, and you should just use the legacy one.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
jpd236: We'd need to use something different than a ThreadPoolExecutor in that case. ScheduledThreadPoolExecutor should work.

(That said, this may not be a problem for now if we drop the timing constraint, although we may want to keep the API as requiring a ScheduledExecutorService so that we'll have that option in the future).

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
sphill99: Currently we only use version 3.4.2 for Roboelectric dependency. To use that, I think we'd need to use a newer version. Would that be fine? 
Using MoreExecutors.directExecutor() won't work since that provides an Executor and not an ExecutorService. Also, we can use RoboExexutorService instead, but I don't think that runs the tasks right away.

sphill99:
I addressed most comments in the last commit. I shared a lot of code between BasicNetwork and BasicAsyncNetwork in NetworkUtility. I also made it so that AsyncNetwork requires an AsyncHttpStack, so now there is no blocking (unless the input stream needs to be read). I also switched back, so that the request is retried immediately right now. 

A few main comments:
1) I stopped using an actual executor in unit tests, but to use InlineExecutorService, I think we'd need to update to a newer Roboelectric version.
2) Currently, there is no way to wait before retrying again. I'm a little confused on how to move forward with this.
3) I shared a lot of code, and it currently obfuscates what's going on a little bit. Wasn't sure if this was the best way to share the code. If so, I'll add descriptive comments to make it more clear what's happening.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
sphill99: I'm confused by what you mean here?

jpd236: I'm suggesting that these methods that relate to header manipulation could be added to HttpHeaderParser instead - this class already has a bunch of utilities around headers, so it seems like we might want to try to keep those together.

sphill99: oh that makes a lot more sense. I thought you were saying this could be done in that class.

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
sphill99: MoreExecutors.newDirectExecutorService() seems like a good option, and works.

src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -169,70 +137,32 @@ public NetworkResponse performRequest(Request<?> request) throws VolleyError {
...
-                    if (statusCode == HttpURLConnection.HTTP_UNAUTHORIZED
-                            || statusCode == HttpURLConnection.HTTP_FORBIDDEN) {
-                        attemptRetryOnException(
-                                "auth", request, new AuthFailureError(networkResponse));
-                    } else if (statusCode >= 400 && statusCode <= 499) {
-                        // Don't retry other client errors.
-                        throw new ClientError(networkResponse);
-                    } else if (statusCode >= 500 && statusCode <= 599) {
-                        if (request.shouldRetryServerErrors()) {
-                            attemptRetryOnException(
```
sophiahaim, line 207: Remove the #attemptRetryOnException method from this class since it's no longer used

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,224 @@
...
+            throw e;
+        }
+        request.addMarker(String.format("%s-retry [timeout=%s]", logPrefix, oldTimeout));
+        if (network instanceof AsyncNetwork) {
+            if (callback == null) {
+                throw new VolleyError("No callback provided");
+            }
+            ((AsyncNetwork) network).performRequest(request, callback);
+        } else {
+            network.performRequest(request);
```
sophiahaim, line 147: I don't see this in the original BasicNetwork#attemptRetryOnException method. Why was it added here? (or maybe I'm missing where this is called in the original implementation)

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+
+/** A network performing Volley requests over an {@link HttpStack}. */
+public class BasicAsyncNetwork extends AsyncNetwork {
+
+    protected final ByteArrayPool mPool;
+
+    protected ExecutorService mBlockingExecutor;
```
sophiahaim, line 42: Now that you set this in AsyncNetwork, I don't think it really makes sense to have it as a class variable here. I think you should just get it (#getBlockingExecutor) in each of the methods that use it (and just set a local var)

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+                    request,
+                    callback,
+                    exception,
+                    requestStartMs,
+                    httpResponse,
+                    responseContents,
+                    this);
+        } catch (VolleyError volleyError) {
+            callback.onError(volleyError);
+        }
```
sophiahaim, line 140: You need a callback in the non-VolleyError case

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
sphill99: One thing I was unclear on: does implementing these methods directly in AsyncNetwork/AsyncHttpStack solve this issue? Like, would we not need the RestrictTo.Scopt.SUBCLASSES part, since we're doing that?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutorService;
+
+/** A network performing Volley requests over an {@link HttpStack}. */
+public class BasicAsyncNetwork extends AsyncNetwork {
+
+    protected final ByteArrayPool mPool;
```
jpd236, line 40: private - do subclasses of BasicAsyncNetwork need direct access to the pool? not sure there's a use case there at the moment.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+    private void onRequestSucceeded(
+            final Request<?> request,
+            final long requestStartMs,
+            final HttpResponse httpResponse,
+            final OnRequestComplete callback) {
+        final int statusCode = httpResponse.getStatusCode();
+        final List<Header> responseHeaders = httpResponse.getHeaders();
+        // Handle cache validation.
+        if (statusCode == HttpURLConnection.HTTP_NOT_MODIFIED) {
+            callback.onSuccess(
```
jpd236, line 64: missing short-circuit return, I think.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,224 @@
...
+                            + "[rc=%d], [retryCount=%s]",
+                    request,
+                    requestLifetime,
+                    responseContents != null ? responseContents.length : "null",
+                    statusCode,
+                    request.getRetryPolicy().getCurrentRetryCount());
+        }
+    }
+
+    static NetworkResponse getNetworkResponse(
```
jpd236, line 65: This method name is too generic for what it actually does - return a response for the HTTP_NOT_MODIFIED case. Perhaps getNotModifiedNetworkResponse.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,224 @@
...
+            int statusCode, Request<?> request, long requestStart, List<Header> responseHeaders) {
```
jpd236, line 66: statusCode seems to be unused.

jpd236, line 66: nit: rather than passing in the start time and calculating the duration yourself, I would suggest that the caller be responsible for passing in the duration. Passing in a start time is a more complex API because the time scale isn't intuitively obvious and easier to get wrong - how do you know/enforce it's based on elapsed realtime?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+                    statusCode,
+                    httpResponse,
+                    request,
+                    callback,
+                    responseHeaders,
+                    responseContents);
+            return;
+        }
+
+        final InputStream inputStream = httpResponse.getContent();
```
jpd236, line 90: Maybe add an explanatory comment here - something like:

"The underlying AsyncHttpStack does not support asynchronous reading of the response into a byte array, so we need to submit a blocking task to copy the response from the InputStream instead."

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+        Runnable run =
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        byte[] finalResponseContents = new byte[0];
```
jpd236, line 95: It shouldn't be necessary to initialize this - it's probably because of the missing short circuit below.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+                        try {
+                            finalResponseContents =
+                                    NetworkUtility.inputStreamToBytes(
+                                            inputStream, httpResponse.getContentLength(), mPool);
+                        } catch (IOException e) {
+                            onRequestFailed(
```
jpd236, line 101: need to return here to avoid calling onResponseRead

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+            callback.onError(volleyError);
+        }
+    }
+
+    @Override
+    public void performRequest(final Request<?> request, final OnRequestComplete callback) {
+        mBlockingExecutor = getBlockingExecutor();
+        if (mBlockingExecutor == null) {
+            throw new IllegalStateException(
+                    "mBlockingExecuter should be set before making a request");
```
jpd236, line 148: nit: s/should/must

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+            Request<?> request,
+            OnRequestComplete callback,
+            List<Header> responseHeaders,
+            byte[] responseContents) {
+        // if the request is slow, log it.
+        long requestLifetime = SystemClock.elapsedRealtime() - requestStartMs;
+        logSlowRequests(requestLifetime, request, responseContents, statusCode);
+
+        if (statusCode < 200 || statusCode > 299) {
+            onRequestFailed(
```
jpd236, line 195: missing short-circuit return. Is it possible to unit test for this and/or do a deeper look for cases where this is happening?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+        callback.onSuccess(
+                new NetworkResponse(
+                        statusCode,
+                        responseContents,
+                        /* notModified= */ false,
+                        SystemClock.elapsedRealtime() - requestStartMs,
+                        responseHeaders));
+    }
+
+    public static class Builder {
```
jpd236, line 213: Javadoc for this and public methods

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+            mAsyncStack = httpStack;
+            mPool = null;
+        }
+
+        public Builder setPool(ByteArrayPool pool) {
+            mPool = pool;
+            return this;
+        }
+
+        public BasicAsyncNetwork build() {
```
jpd236, line 228: throw is mAsyncStack is null?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+                        statusCode,
+                        responseContents,
+                        /* notModified= */ false,
+                        SystemClock.elapsedRealtime() - requestStartMs,
+                        responseHeaders));
+    }
+
+    public static class Builder {
+        private static final int DEFAULT_POOL_SIZE = 4096;
+        AsyncHttpStack mAsyncStack;
```
jpd236, line 215: private

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,235 @@
...
+        ByteArrayPool mPool;
```
jpd236, line 216: private

src/main/java/com/android/volley/toolbox/BasicNetwork.java
```diff
@@ -169,70 +137,32 @@ public NetworkResponse performRequest(Request<?> request) throws VolleyError {
...
-                    statusCode,
-                    request.getRetryPolicy().getCurrentRetryCount());
+    /**
+     * Converts Headers[] to Map&lt;String, String&gt;.
+     *
+     * @deprecated Should never have been exposed in the API. This method may be removed in a future
+     *     release of Volley.
+     */
+    @Deprecated
+    static Map<String, String> convertHeaders(Header[] headers) {
```
jpd236, line 160: this was protected before - it should remain protected.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -0,0 +1,224 @@
...
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.SocketTimeoutException;
+import java.util.List;
+
+/**
+ * Utility class for methods that are shared between {@link BasicNetwork} and {@link
+ * BasicAsyncNetwork}
+ */
+public class NetworkUtility {
```
jpd236, line 47: mark final, add private constructor to prevent instantiation of this utility class

src/main/java/com/android/volley/toolbox/NetworkUtility.java
jpd236: Yeah,  I don't think this quite works. BasicNetwork works today by running the request in a while(true) loop until this method throws; calling performRequest directly is sort of counter to that since it makes the loop redundant.

I think we either do the same thing in the async network (except we need to deal with asynchronicity) or do something like return a boolean here and let the caller inspect the result to determine if a retry is necessary, rather than passing the network in and making a call back out to it.

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
```diff
@@ -0,0 +1,435 @@
...
+import java.util.concurrent.ExecutorService;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(sdk = 16)
```
jpd236, line 65: Is this needed? What for?

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: is it better to throw, or add the @NonNull annotation? I get a warning if I do both.

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
sphill99: That is no longer needed. I had tried something that required a higher API, but didn't use it.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: Are you asking whether it's possible to unit test whether we return correctly? The only thing I can think of is verifying that methods returning are never called. But that relies on us knowing where they all are. I think just making different types of requests in a test app and making sure they don't make two callback calls is the best way to do it.

jpd236: Right, my hope is that it would be possible to wire up a request with a mock stack and validate that the expected callback is invoked exactly once and that the other callback methods are not invoked at all.

sphill99:
I believe I made all the appropriate fixes from the comments. I no longer pass in the network or callback into handleException. It either throws the exception, or does nothing. If no exception is thrown, the request is retried (by calling performRequest in async version, and by looping again in non-async version). I believe all the short circuit return issues have been addressed. Will comb through again just to make sure. 

One thing to note. I couldn't make the setter methods in AsyncHttpStack protected, since AsyncNetwork (which calls the setters), is not in the same package. Currently they're public, but we could move AsyncHttpStack outside of the toolbox.

src/main/java/com/android/volley/toolbox/NetworkUtility.java
sphill99: In every case, either an error should be thrown (/passed through callback) or the request should be retried. Because of this, I just catch the error, call the callback, and return, or perform the request. Also, now BasicNetwork works the same as it originally did (either error is thrown to get out of while loop, or it's looped again).

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: Now that the callback is called outside of the handleException method, I think it's clearer to the reader that it's always being thrown.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: okay, I'll work on that. I think that shouldn't be too hard actually.

sphill99:
Just added testing that makes sure that the callback is only ever called once. Added a mock callback, and then verified the number of times onSuccess and onError are called. I'm not super familiar with Mockito though, so not sure if this is the best way to do this.

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,124 @@
...
+        } else if (error.get() != null) {
+            throw error.get();
+        } else {
+            throw new VolleyError("Neither response entry was set");
+        }
+    }
+
+    /**
+     * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
+     * This method must be called before performing any requests if you are using an AsyncHttpStack.
```
sophiahaim, line 95: Remove (since always using an AsyncHttpsStack now)

src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -70,7 +64,7 @@ public void executeRequest(
             final Request<?> request,
             final Map<String, String> additionalHeaders,
             final OnRequestComplete callback) {
-        if (mBlockingExecutor == null || mCallbackExecutor == null) {
+        if (getBlockingExecutor() == null || getNonBlockingExecutor() == null) {
             throw new IllegalStateException("Must set blocking / non-blocking executors");
```
sophiahaim, line 68: super nit, but I think this should be an "and" ("Must set blocking and non-blocking executors")

sphill99: Shouldn't it be that if either of them are null, we throw the exception?

sophiahaim: Yep, I'm just referring to the IllegalStateException method, since both should be set. But yes, the code itself is right (should throw if either are null since both need to be set)

sophiahaim: *the IllegalStateException message

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,124 @@
...
+public abstract class AsyncNetwork implements Network {
+    private final AsyncHttpStack mAsyncStack;
+    private ExecutorService mBlockingExecutor;
+    private ExecutorService mNonBlockingExecutor;
+
+    protected AsyncNetwork(AsyncHttpStack stack) {
+        mAsyncStack = stack;
+    }
+
+    public interface OnRequestComplete {
```
jpd236, line 35: javadoc

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,124 @@
...
+     * This method sets the blocking executor to be used by the network and stack for potentially
+     * blocking tasks. This method must be called before performing any requests.
+     */
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
+    public void setBlockingExecutor(ExecutorService executor) {
+        mBlockingExecutor = executor;
+        mAsyncStack.setBlockingExecutor(executor);
+    }
+
+    protected ExecutorService getBlockingExecutor() {
```
jpd236, line 113: javadoc - in particular, what sorts of tasks should subclasses use this executor for?

also for getNonBlockingExecutor and getHttpStack.

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,124 @@
...
+            throw new VolleyError("Neither response entry was set");
+        }
+    }
+
+    /**
+     * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
+     * This method must be called before performing any requests.
+     */
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
+    public void setNonBlockingExecutorForStack(ExecutorService executor) {
```
jpd236, line 98: "setNonBlockingExecutor"? Not sure what the "ForStack" is for here since it's no different from what setBlockingExecutor is doing. This executor could be used by the network too.

src/main/java/com/android/volley/AsyncNetwork.java
```diff
@@ -0,0 +1,124 @@
...
+            return response.get();
+        } else if (error.get() != null) {
+            throw error.get();
+        } else {
+            throw new VolleyError("Neither response entry was set");
+        }
+    }
+
+    /**
+     * This method sets the non blocking executor to be used by the stack for non-blocking tasks.
```
jpd236, line 94: by the network and stack?

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
```diff
@@ -57,15 +59,27 @@ public abstract void executeRequest(
...
      * tasks. This method must be called before executing any requests.
      */
-    @RestrictTo({RestrictTo.Scope.SUBCLASSES, RestrictTo.Scope.LIBRARY_GROUP})
-    public abstract void setBlockingExecutor(ExecutorService executor);
+    @RestrictTo({RestrictTo.Scope.LIBRARY_GROUP})
+    public void setBlockingExecutor(ExecutorService executor) {
+        mBlockingExecutor = executor;
+    }
+
+    protected ExecutorService getBlockingExecutor() {
```
jpd236, line 76: javadoc for this and getNonBlockingExecutor

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
```diff
@@ -0,0 +1,244 @@
...
+                                    NetworkUtility.inputStreamToBytes(
+                                            inputStream, httpResponse.getContentLength(), mPool);
+                        } catch (IOException e) {
+                            onRequestFailed(
+                                    request, callback, e, requestStartMs, httpResponse, null);
+                            return;
+                        } catch (ServerError serverError) {
+                            // This should never happen since we already check if inputStream is
+                            // null
+                            throw new RuntimeException(serverError);
```
jpd236, line 109: perhaps IllegalStateException - or consider changing inputStreamToBytes not to throw ServerError and moving that logic into BasicNetwork if it's unused in BasicAsyncNetwork's flow.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
jpd236: OK - we haven't used `@NonNull` much before, but I suppose it's best here because it can be caught at compile time, whereas throwing is only caught at runtime. I'm not 100% convinced b/c it depends on Android Studio tooling and other developers might use other tools that don't detect the problem, but it'll crash at runtime anyway.

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
```diff
@@ -0,0 +1,509 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
+public class BasicAsyncNetworkTest {
+
+    CompletableFuture<NetworkResponse> future = new CompletableFuture<>();
```
jpd236, line 67: nit: private

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
```diff
@@ -0,0 +1,509 @@
...
+    @Mock private RetryPolicy mMockRetryPolicy;
+    @Mock AsyncNetwork.OnRequestComplete mockCallback;
```
jpd236, line 69: private

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
```diff
@@ -0,0 +1,509 @@
...
+        entry.etag = "foobar";
+        entry.lastModified = 1503102002000L;
+        request.setCacheEntry(entry);
+        httpNetwork.performRequest(request, mockCallback);
+        verify(mockCallback, times(1)).onSuccess(any(NetworkResponse.class));
+        verify(mockCallback, never()).onError(any(VolleyError.class));
+        reset(mMockRetryPolicy, mockCallback);
+    }
+
+    @Test(expected = IllegalStateException.class)
```
jpd236, line 453: Can we use Assert.assertThrows without adding a new dependency? If so, that's preferable since it more tightly describes the code expected to fail, rather than just letting any IllegalStateException thrown anywhere inside this test cause the test to pass.

src/main/java/com/android/volley/toolbox/BasicAsyncNetwork.java
sphill99: I think handling it in BasicNetwork is better, and then not dealing with the ServerError in BasicAsyncNetwork.

src/test/java/com/android/volley/toolbox/BasicAsyncNetworkTest.java
sphill99: I don't think so, since that's only in JUnit5. I'm not aware of another way to test that that specific code causes it.

src/main/java/com/android/volley/AsyncNetwork.java
sphill99: My reasoning was that getNonBlockingExecutor() isn't called at all by BasicAsyncNetwork. But you're right that it could be used by the network.

sphill99:
I believe I've now addressed all comments. The only thing I didn't change was using assertThrows, but I don't believe we are able to since it requires JUnit 5.

src/main/java/com/android/volley/toolbox/AsyncHttpStack.java
sophiahaim: I think this javadoc wasn't included in your last commit, please add, thanks!

PR #359, opened 2020-08-17
Fixed the null context inside CronetHtppStack.Builder class, and adde…

leoggoes:
…d default executors in case none is set.

### DIFF
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -36,7 +36,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
+import java.util.concurrent.*;
+
 import org.chromium.net.CronetEngine;
 import org.chromium.net.CronetException;
 import org.chromium.net.UploadDataProvider;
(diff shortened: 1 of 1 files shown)


jpd236:
@sphill99 FYI

Thanks for sending this. The Cronet/Async HTTP stack is very much under construction and not suitable for use, and we're actively iterating on it, so at the moment we're not looking for contributions and do know that there are outstanding issues and gaps. We would definitely not recommend using or developing against it at this time unless/until it makes it into an official release.

Will leave this open to make sure we have resolved the issue being reported in some form before we make this API more broadly available.

leoggoes:
@jpd236 I understand. Those are just minor changes to make things running. 
Using volley with this cronet early version less than halved the response time for my requests. 
If I encounter any issues, i'll report them asap.

jpd236:
OK - looks more closely at this, I think there's no need for further work to address the issues raised in this PR. The Context NPE looks to have been fixed. For the executors, at the moment, the intended way to use CronetHttpStack is as part of AsyncRequestQueue/BasicAsyncNetwork, and AsyncRequestQueue will provide default executors if none are provided. Though we may want to think a bit about whether it's benefiting us to have AsyncHttpStack extend BaseHttpStack, since it's not really a drop-in replacement; the executors won't be populated, and populating them requires calling library-internal methods.

jpd236:
And you can consult https://github.com/google/volley/wiki/Asynchronous-Volley for how to use AsyncRequestQueue/BasicAsyncNetwork.

PR #360, opened 2020-08-18
Created builder for DiskBasedAsyncCache

sphill99:
I realized that all the same constructors for DiskBasedCache weren't there for DiskBasedAsyncCache. I added a builder to set the FileSupplier and max size. Could also just use the same telescoping scheme as in DiskBasedCache, but figured this was consistent with the rest of the async implementations.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -411,8 +411,8 @@ public class DiskBasedAsyncCache extends AsyncCache {
      * the setters.
      */
     public static class Builder {
-        private FileSupplier rootDirectorySupplier;
-        private File rootDirectory;
+        @Nullable private FileSupplier rootDirectorySupplier;
+        @Nullable private File rootDirectory;
         private int maxCacheSizeInBytes;
 
         public Builder() {
--- a/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
+++ b/src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
@@ -59,16 +59,10 @@ public class DiskBasedAsyncCache extends AsyncCache {
     /**
      * Constructs an instance of the DiskBasedAsyncCache at the specified directory.
      *
-     * @param rootDirectory The root directory of the cache.
+     * @param rootDirectorySupplier The root directory supplier of the cache.
      */
-    public DiskBasedAsyncCache(final File rootDirectory, int maxCacheSizeInBytes) {
-        mRootDirectorySupplier =
-                new FileSupplier() {
-                    @Override
-                    public File get() {
-                        return rootDirectory;
-                    }
-                };
+    private DiskBasedAsyncCache(final FileSupplier rootDirectorySupplier, int maxCacheSizeInBytes) {
+        mRootDirectorySupplier = rootDirectorySupplier;
         mMaxCacheSizeInBytes = maxCacheSizeInBytes;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -411,4 +405,67 @@ private void deleteFileAndInvokeCallback(
         mTotalSize = DiskBasedCacheUtility.removeEntry(key, mTotalSize, mEntries);
         callback.onGetComplete(null);
     }
+
+    /**
+     * Builder is used to build an instance of {@link DiskBasedAsyncCache} from values configured by
+     * the setters.
+     */
+    public static class Builder {
+        private FileSupplier rootDirectorySupplier;
```
sophiahaim, line 414: nit: Mark these as @Nullable

sphill99: I was getting an error when I did that, because it wanted the setter argument to be nullable as well. Should both be nullable, or is it better to leave the warning?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -59,16 +59,10 @@
...
      */
-    public DiskBasedAsyncCache(final File rootDirectory, int maxCacheSizeInBytes) {
-        mRootDirectorySupplier =
-                new FileSupplier() {
-                    @Override
-                    public File get() {
-                        return rootDirectory;
-                    }
-                };
+    private DiskBasedAsyncCache(final FileSupplier rootDirectorySupplier, int maxCacheSizeInBytes) {
```
jbir789, line 64: Do we really need to mark the FileSupplier parameter `final`? I feel like it isn't needed, especially as a constructor parameter: it's unlikely the variable will be re-assigned within the constructor.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -411,4 +405,67 @@ private void deleteFileAndInvokeCallback(
...
+     */
+    public static class Builder {
+        @Nullable private FileSupplier rootDirectorySupplier;
+        @Nullable private File rootDirectory;
+        private int maxCacheSizeInBytes;
+
+        public Builder() {
+            rootDirectorySupplier = null;
+            rootDirectory = null;
+            maxCacheSizeInBytes = -1;
```
jbir789, line 421: nit: what do you think of initializing those values at the declaration? This would be more concise, as we could remove the constructor

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -411,4 +405,67 @@ private void deleteFileAndInvokeCallback(
...
+        }
+
+        /**
+         * Sets the root directory of the cache. Must be called if {@link
+         * Builder#setRootDirectorySupplier(FileSupplier)} is not.
+         */
+        public Builder setRootDirectory(File file) {
```
jbir789, line 428: nit: should the parameter be called rootDirectory for consistency?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -411,4 +405,67 @@ private void deleteFileAndInvokeCallback(
...
+            this.rootDirectory = file;
+            return this;
+        }
+
+        /**
+         * Sets the root directory supplier of the cache. Must be called if {@link
+         * Builder#setRootDirectory(File)} is not.
+         */
+        public Builder setRootDirectorySupplier(FileSupplier fileSupplier) {
```
jbir789, line 437: Same here: `rootDirectorySupplier`

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
```diff
@@ -411,4 +405,67 @@ private void deleteFileAndInvokeCallback(
...
+        }
+
+        /** Builds a DiskBasedAsyncCache from the provided parameters. */
+        public DiskBasedAsyncCache build() {
+            if (rootDirectory == null && rootDirectorySupplier == null) {
+                throw new IllegalArgumentException("Must set either file or supplier");
+            }
+            if (maxCacheSizeInBytes == -1) {
+                maxCacheSizeInBytes = DiskBasedCacheUtility.DEFAULT_DISK_USAGE_BYTES;
+            }
```
jbir789, line 458: What if the default value was `DEFAULT_DISK_USAGE_BYTES` instead of `-1`?

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Yeah, not sure why that's there. Don't think it's necessary.

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: I like that

src/main/java/com/android/volley/toolbox/DiskBasedAsyncCache.java
sphill99: Oh yep, that saves us from writing this code block in build().

PR #361, opened 2020-08-19
AsyncRequestQueue implementation + some tests

sphill99:
Implementation for AsyncRequestQueue. Overrides start / stop / sendRequestOverNetwork. add was changed in RequestQueue to now invoke beginTask which is overridden in AsyncRequestQueue. WaitingRequestManager was to its own class so that both CacheDispatcher and AsyncRequestQueue can use it. (From what I can tell, that doesn't break any APIs).

Currently it handles either a cache or asyncCache and branches depending on which is being used.

One thing that I noticed was different than the current requestQueue is that start() must be called before adding requests. This is because the executors are null until we call start(). Not sure if this matters, and if it does, where we should be initializing the executors.

I tried to copy the RequestQueue.Test to AsyncRequestQueue.Test. Not sure if there are other unit tests we should be doing for this. I also tested on test app and it seems to be working based on that.

I didn't add the newAsyncRequestQueue() method in Volley.java into this PR. Wasn't sure if we should wait for that, but I have code written for that if we want to add that too.

### DIFF
--- a/src/main/java/com/android/volley/AsyncRequestQueue.java
+++ b/src/main/java/com/android/volley/AsyncRequestQueue.java
@@ -1,0 +1,558 @@
+/*
+ * Copyright (C) 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.android.volley;
+
+import android.os.Handler;
+import android.os.Looper;
+import android.os.SystemClock;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.android.volley.AsyncCache.OnGetCompleteCallback;
+import com.android.volley.AsyncNetwork.OnRequestComplete;
+import com.android.volley.Cache.Entry;
+import com.android.volley.toolbox.ThrowingCache;
+import java.net.HttpURLConnection;
+import java.util.Comparator;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class AsyncRequestQueue extends RequestQueue {
+    /** Default number of blocking threads to start. */
+    private static final int DEFAULT_BLOCKING_THREAD_POOL_SIZE = 4;
+
+    /** AsyncCache used to retrieve and store responses. Null indicates use of blocking Cache. */
+    @Nullable private final AsyncCache mAsyncCache;
+
+    /** AsyncNetwork used to perform nework requests. */
+    private final AsyncNetwork mNetwork;
+
+    /**
+     * Cache used to retrieve and store responses. Only use if an AsyncCache cannot be provided due
+     * to lower API.
+     */
+    private final Cache mCache;
+
+    /** Executor for non-blocking tasks. */
+    private ExecutorService mNonBlockingExecutor;
+
+    /** Blocking queue to be used to created Executors */
+    private static final PriorityBlockingQueue<Runnable> mBlockingQueue =
+            new PriorityBlockingQueue<>(
+                    11,
+                    new Comparator<Runnable>() {
+                        @Override
+                        public int compare(Runnable r1, Runnable r2) {
+                            // Vanilla runnables are prioritized first, then RequestTasks are
+                            // ordered
+                            // by the underlying Request.
+                            if (r1 instanceof RequestTask) {
+                                if (r2 instanceof RequestTask) {
+                                    return ((RequestTask<?>) r1).compareTo(((RequestTask<?>) r2));
+                                }
+                                return 1;
+                            }
+                            return r2 instanceof RequestTask ? -1 : 0;
+                        }
+                    });
+
+    /**
+     * Executor for blocking tasks.
+     *
+     * <p>Some tasks in handling requests may not be easy to implement in a non-blocking way, such
+     * as reading or parsing the response data. This executor is used to run these tasks.
+     */
+    private ExecutorService mBlockingExecutor;
+
+    /**
+     * This interface may be used by advanced applications to provide custom executors according to
+     * their needs. Apps must create ExecutorServices dynamically given a blocking queue rather than
+     * providing them directly so that Volley can provide a PriorityQueue which will prioritize
+     * requests according to Request#getPriority.
+     */
+    private ExecutorFactory mExecutorFactory;
+
+    /** Manage list of waiting requests and de-duplicate requests with same cache key. */
+    private final WaitingRequestManager mWaitingRequestManager;
+
+    /**
+     * Sets all the variables, but processing does not begin until {@link #start()} is called.
+     *
+     * @param cache to use for persisting responses to disk. If an AsyncCache was provided, then
+     *     this will be a {@link ThrowingCache}
+     * @param network to perform HTTP requests
+     * @param asyncCache to use for persisting responses to disk. May be null to indicate use of
+     *     blocking cache
+     * @param responseDelivery interface for posting responses and errors
+     * @param executorFactory Interface to be used to provide custom executors according to the
+     *     users needs.
+     */
+    private AsyncRequestQueue(
+            Cache cache,
+            AsyncNetwork network,
+            @Nullable AsyncCache asyncCache,
+            ResponseDelivery responseDelivery,
+            @Nullable ExecutorFactory executorFactory) {
+        super(cache, network, 0, responseDelivery);
+        mAsyncCache = asyncCache;
+        mCache = cache;
+        mNetwork = network;
+        mExecutorFactory = executorFactory;
+        mWaitingRequestManager = new WaitingRequestManager(this);
+    }
+
+    /** Sets the executors and initializes the cache. */
+    @Override
+    public void start() {
+        stop(); // Make sure any currently running threads are stopped
+
+        // Create blocking / non-blocking executors and set them in the network and stack.
+        mNonBlockingExecutor = mExecutorFactory.createNonBlockingExecutor(mBlockingQueue);
+        mBlockingExecutor = mExecutorFactory.createBlockingExecutor(mBlockingQueue);
+        mNetwork.setBlockingExecutor(mBlockingExecutor);
+        mNetwork.setNonBlockingExecutor(mNonBlockingExecutor);
+
+        mNonBlockingExecutor.execute(
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        // This is intentionally blocking, because we don't want to process any
+                        // requests until the cache is initialized.
+                        if (mAsyncCache != null) {
+                            final CountDownLatch latch = new CountDownLatch(1);
+                            mAsyncCache.initialize(
+                                    new AsyncCache.OnWriteCompleteCallback() {
+                                        @Override
+                                        public void onWriteComplete() {
+                                            latch.countDown();
+                                        }
+                                    });
+                            try {
+                                latch.await();
+                            } catch (InterruptedException e) {
+                                VolleyLog.e(
+                                        e, "Thread was interrupted while initializing the cache.");
+                                Thread.currentThread().interrupt();
+                                throw new RuntimeException(e);
+                            }
+                        } else {
+                            mCache.initialize();
+                        }
+                    }
+                });
+    }
+
+    /** Shuts down and nullifies both executors */
+    @Override
+    public void stop() {
+        if (mNonBlockingExecutor != null) {
+            mNonBlockingExecutor.shutdownNow();
+            mNonBlockingExecutor = null;
+        }
+        if (mBlockingExecutor != null) {
+            mBlockingExecutor.shutdownNow();
+            mBlockingExecutor = null;
+        }
+    }
+
+    /** Begins the request by sending it to the Cache or Network. */
+    @Override
+    <T> void beginRequest(Request<T> request) {
+        // If the request is uncacheable, send it over the network.
+        if (request.shouldCache()) {
+            if (mAsyncCache != null) {
+                mNonBlockingExecutor.execute(new CacheTask<>(request));
+            } else {
+                mBlockingExecutor.execute(new CacheTask<>(request));
+            }
+        } else {
+            sendRequestOverNetwork(request);
+        }
+    }
+
+    @Override
+    <T> void sendRequestOverNetwork(Request<T> request) {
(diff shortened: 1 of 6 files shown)


src/test/java/com/android/volley/AsyncRequestQueueTest.java
```diff
@@ -0,0 +1,159 @@
...
+import com.android.volley.toolbox.StringRequest;
+import com.android.volley.utils.ImmediateResponseDelivery;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+/** Unit tests for RequestQueue, with all dependencies mocked out */
```
jpd236, line 37: nit: AsyncRequestQueue

src/test/java/com/android/volley/AsyncRequestQueueTest.java
```diff
@@ -0,0 +1,159 @@
...
+        verify(req4, never()).cancel(); // A added after cancel not cancelled
+        queue.stop();
+    }
+
+    @Test
+    public void add_notifiesListener() throws Exception {
+        RequestQueue.RequestEventListener listener = mock(RequestQueue.RequestEventListener.class);
+        AsyncRequestQueue queue =
+                new AsyncRequestQueue.Builder(mMockNetwork)
+                        .setCache(new NoCache())
```
jpd236, line 88: should we define an Async equivalent of NoCache and use it here to exercise the Async flow?

src/test/java/com/android/volley/AsyncRequestQueueTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        mDelivery = new ImmediateResponseDelivery();
+        initMocks(this);
+    }
+
+    @Test
+    public void cancelAll_onlyCorrectTag() throws Exception {
+        AsyncRequestQueue queue =
```
jpd236, line 53: We shouldn't rely on the default executors for tests - ideally we'd use controlled ones, advance them until they reach a terminal state as needed, and then verify the result.

jpd236, line 53: how much of this queue creation could be done in setUp()? It seems like we end up creating basically the same queue for each test.

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,558 @@
...
+    private ExecutorService mNonBlockingExecutor;
+
+    /** Blocking queue to be used to created Executors */
+    private static final PriorityBlockingQueue<Runnable> mBlockingQueue =
+            new PriorityBlockingQueue<>(
+                    11,
+                    new Comparator<Runnable>() {
+                        @Override
+                        public int compare(Runnable r1, Runnable r2) {
+                            // Vanilla runnables are prioritized first, then RequestTasks are
```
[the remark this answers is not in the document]
sphill99: Okay, that makes sense. So then we'd have CacheGetTask, CachePutTask, CacheParseTask, NetworkTask, NetworkParseTask, and any others I'm missing? Just want to make sure that's the general idea.

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,558 @@
...
+            if (mRequest.isCanceled()) {
+                mRequest.finish("network-discard-cancelled");
+                mRequest.notifyListenerResponseNotUsable();
+                return;
+            }
+
+            final long startTimeMs = SystemClock.elapsedRealtime();
+            mRequest.addMarker("network-queue-take");
+
+            // TODO: Figure out what to do with traffic stats tags. Can this be pushed to the
```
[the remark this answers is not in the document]
sphill99: I have it in a document as well. I figured it can't hurt to keep the TODO in there, but lmk if I should remove it.

src/test/java/com/android/volley/AsyncRequestQueueTest.java
sphill99: Yeah, I think that would be a good idea. Would that just look like a method that calls the callback with null?

src/test/java/com/android/volley/AsyncRequestQueueTest.java
sphill99: Yeah, I think it can entirely be done. I kept the same format from RequestQueue.test

src/main/java/com/android/volley/toolbox/ThrowingCache.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.android.volley.toolbox;
+
+import com.android.volley.Cache;
+
+/** A cache that thorws an error if a method is called. */
+public class ThrowingCache implements Cache {
```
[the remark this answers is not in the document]
sphill99: I think we can only do that if we move it to the same package, and make it package-private. Unless the restrictTo annotation would work here.

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,558 @@
...
+                                            latch.countDown();
+                                        }
+                                    });
+                            try {
+                                latch.await();
+                            } catch (InterruptedException e) {
+                                VolleyLog.e(
+                                        e, "Thread was interrupted while initializing the cache.");
+                                Thread.currentThread().interrupt();
+                                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
sphill99: My understanding was that there'd be an NPE and the whole app would crash. but I can try to test that and see.

sphill99:
I've either addressed or responded to all comments. I'm still working on converting vanilla runnables on network/stack layer to RequestTasks. I'm also looking into the case where a thread crashes.

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,581 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * An asynchronous request dispatch queue.
+ *
+ * <p>Calling {@link #add(Request)} will submit a task to the non-blocking executor. Once completed,
```
jpd236, line 42: I guess this is copied from RequestQueue, but it's not really correct, because it assumes you're using the default ResponseDelivery implementation. I think the non-blocking executor is also a bit of an implementation detail. Perhaps just something like:

```Add requests to the queue with {@link #add(Request)}. Once completed, responses will be delivered on the main thread (unless a custom {@link ResponseDelivery} has been provided).```

File: src/main/java/com/android/volley/AsyncRequestQueue.java [hunk 9]
```diff
@@ -0,0 +1,581 @@
...
+                            mBlockingExecutor.execute(
+                                    new NetworkParseTask<>(mRequest, networkResponse));
+                        }
+
+                        @Override
+                        public void onError(final VolleyError volleyError) {
+                            volleyError.setNetworkTimeMs(
+                                    SystemClock.elapsedRealtime() - startTimeMs);
+                            mBlockingExecutor.execute(
+                                    new Runnable() {
```
jpd236, line 352: this can be a RequestTask (e.g. ParseErrorTask)

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,581 @@
...
+            // TODO: Only update cache metadata instead of entire
+            // record for 304s.
+            if (mRequest.shouldCache() && response.cacheEntry != null) {
+                if (mAsyncCache != null) {
+                    mNonBlockingExecutor.execute(new CachePutTask<>(mRequest, response));
+                } else {
+                    mBlockingExecutor.execute(new CachePutTask<>(mRequest, response));
+                }
+            } else {
+                finishRequest(mRequest, response, false);
```
jpd236, line 390: nit: label this boolean argument in all callers (```/* cached= */ <value>```)

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,581 @@
...
+     * requests according to Request#getPriority.
+     */
+    public interface ExecutorFactory {
+        ExecutorService createNonBlockingExecutor(BlockingQueue<Runnable> taskQueue);
+
+        ExecutorService createBlockingExecutor(BlockingQueue<Runnable> taskQueue);
+    }
+
+    /** Provides a BlockingQueue to be used to create executors. */
+    private PriorityBlockingQueue<Runnable> getBlockingQueue() {
```
jpd236, line 446: this can be static, right?

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,581 @@
...
+        return new PriorityBlockingQueue<>(
+                /* initialCapacity= */ 11,
+                new Comparator<Runnable>() {
+                    @Override
+                    public int compare(Runnable r1, Runnable r2) {
+                        // Vanilla runnables are prioritized first, then RequestTasks are
+                        // ordered
+                        // by the underlying Request.
```
jpd236, line 454: nit: move to previous line? weird line-breaking in this comment

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,581 @@
...
+            return this;
+        }
+
+        /** Provides a default ExecutorFactory to use, if one is never set. */
+        private ExecutorFactory getDefaultExecutorFactory() {
+            return new ExecutorFactory() {
+                @Override
+                public ExecutorService createNonBlockingExecutor(
+                        BlockingQueue<Runnable> taskQueue) {
+                    return new ThreadPoolExecutor(
```
jpd236, line 525: nit: you could reduce some code duplication here with a helper method newThreadPoolExecutor that takes the maximum pool size and thread name suffix

src/main/java/com/android/volley/toolbox/NoAsyncCache.java
```diff
@@ -0,0 +1,36 @@
+package com.android.volley.toolbox;
+
+import com.android.volley.AsyncCache;
+import com.android.volley.Cache;
+
+public class NoAsyncCache extends AsyncCache {
```
jpd236, line 6: nit: add Javadoc. ("An AsyncCache that doesn't cache anything.")

src/main/java/com/android/volley/WaitingRequestManager.java
```diff
@@ -0,0 +1,157 @@
...
+     *       key.
+     *   <li>get(cacheKey) returns waiting requests for the given cache key. The in flight request
+     *       is <em>not</em> contained in that list. Is null if no requests are staged.
+     * </ul>
+     */
+    private final Map<String, List<Request<?>>> mWaitingRequests = new HashMap<>();
+
+    private final ResponseDelivery mResponseDelivery;
+
+    private final RequestQueue mRequestQueue;
```
sophiahaim, line 41: Make all these @Nullable and add javadoc explaining when they are null/non-null

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,581 @@
...
+     * @param responseDelivery interface for posting responses and errors
+     * @param executorFactory Interface to be used to provide custom executors according to the
+     *     users needs.
+     */
+    private AsyncRequestQueue(
+            Cache cache,
+            AsyncNetwork network,
+            @Nullable AsyncCache asyncCache,
+            ResponseDelivery responseDelivery,
+            @Nullable ExecutorFactory executorFactory) {
```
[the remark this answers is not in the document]
sphill99: oh you're right. It used to be, but changed it.

src/main/java/com/android/volley/WaitingRequestManager.java
sphill99: When I do this, I also get warnings for calling methods on the dispatchers. I added a check for non-null, but wasn't sure if this is the cleanest way to do it.

src/main/java/com/android/volley/AsyncRequestQueue.java
sphill99: Okay, that sounds good

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,573 @@
...
+ * An asynchronous request dispatch queue.
+ *
+ * <p>Add requests to the queue with {@link #add(Request)}. Once completed, responses will be
+ * delivered on the main thread (unless a custom {@link ResponseDelivery} has been provided)
+ */
+public class AsyncRequestQueue extends RequestQueue {
+    /** Default number of blocking threads to start. */
+    private static final int DEFAULT_BLOCKING_THREAD_POOL_SIZE = 4;
+
+    /** AsyncCache used to retrieve and store responses. Null indicates use of blocking Cache. */
```
jbir789, line 49: nit: could you have the second sentence on a new line, so the first line is the description of the variable?

also, prefer writing: `{@code null} indicates...`

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,573 @@
...
+            Cache cache,
+            AsyncNetwork network,
+            @Nullable AsyncCache asyncCache,
+            ResponseDelivery responseDelivery,
+            ExecutorFactory executorFactory) {
+        super(cache, network, /* threadPoolSize= */ 0, responseDelivery);
+        mAsyncCache = asyncCache;
+        mNetwork = network;
+        mExecutorFactory = executorFactory;
+        mWaitingRequestManager = new WaitingRequestManager(this);
```
jbir789, line 99: You could initialize this at the declaration.

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,573 @@
...
+    /** Runnable that performs the network request */
+    private class NetworkTask<T> extends RequestTask<T> {
+        NetworkTask(Request<T> request) {
+            super(request);
+        }
+
+        @Override
+        public void run() {
+            // If the request was cancelled already, do not perform the
+            // network request.
```
jbir789, line 315: could this be on the same line as the start of the sentence?

src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -0,0 +1,573 @@
...
+
+        public Builder(AsyncNetwork asyncNetwork) {
+            if (asyncNetwork == null) {
+                throw new IllegalArgumentException("Network cannot be null");
+            }
+            mNetwork = asyncNetwork;
+            mExecutorFactory = null;
+            mResponseDelivery = null;
+            mAsyncCache = null;
+            mCache = null;
```
jbir789, line 477: You can set those at the declaration, and mark the fields as nullable?

src/main/java/com/android/volley/WaitingRequestManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+
+class WaitingRequestManager implements Request.NetworkRequestCompleteListener {
```
jbir789, line 27: nit: can you add javadoc?

PR #367, opened 2020-09-26
Provide a ScheduledExecutorService to AsyncRequestQueue/AsyncNetwork.

jpd236:
This is a subset of #363 which ensures a ScheduledExecutorService is available,
but does not yet define a new RetryPolicy interface. This gives us the
flexibility to introduce timer-based operations later without needing to change
the ExecutorFactory interface (which is now an abstract class to allow for future
additions, should any others be necessary).

### DIFF
--- a/src/main/java/com/android/volley/AsyncRequestQueue.java
+++ b/src/main/java/com/android/volley/AsyncRequestQueue.java
@@ -32,6 +32,8 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -426,15 +437,17 @@ private void finishRequest(Request<?> mRequest, Response<?> response, boolean ca
     }
 
     /**
-     * This interface may be used by advanced applications to provide custom executors according to
+     * This class may be used by advanced applications to provide custom executors according to
      * their needs. Apps must create ExecutorServices dynamically given a blocking queue rather than
      * providing them directly so that Volley can provide a PriorityQueue which will prioritize
      * requests according to Request#getPriority.
      */
-    public interface ExecutorFactory {
```
jbir789, line 434: Why not keeping the interface?

jbir789 (approved):

src/main/java/com/android/volley/AsyncRequestQueue.java
jpd236: This is taken from the Android API guidelines - since we can't use default interface methods in Volley, abstract classes are preferable because it gives us the flexibility to add methods in later versions. Since it seems like there's clearly value in such flexibility (as we're adding a scheduled Executor in this very PR), it seemed like a good time to make it an abstract class as well. I can't think of a concrete example of something we'd be likely to add here now that the scheduled executor is here, but I also don't see much benefit to keeping this an interface since I can't imagine anybody needs multiple inheritance to implement it.

PR #370, opened 2020-09-29
Add cURL logging support to CronetHttpStack.

jpd236:
For debugging purposes, applications may ask Volley to log an
equivalent cURL command for any requests being issued through
this HTTP stack.

Some caveats/notes:

- UrlRequest (and UrlRequest.Builder) don't provide getters for
  the parameters we set on them, and they can't be trivially
  read from the Request either. So we introduce an internal
  container class, CurlLoggedRequestParameters, to store any
  parameters that we want to both propagate to the UrlRequest as
  well as include in cURL logging when enabled. This still isn't
  perfect - Cronet may append additional request headers of its
  own, such as user agent. But this should work as a close
  approximation, and there doesn't appear to be a better way.

- By default, we attempt to redact any auth tokens by redacting
  the values of Authorization and Cookie headers, while providing
  an option to bypass this if needed.

- We have a crude heuristic to detect text vs. binary content so
  that known text content can be logged in plaintext when passed
  to cURL. It's certainly not complete, but for debugging purposes
  it should suffice for now.

- CronetHttpStack has a bug compared to HurlStack where it fails to
  set the Content-Type header when there is a request body. Some test
  output is expected to change once this bug is resolved.

### DIFF
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -17,25 +17,30 @@
 package com.android.volley.cronet;
 
 import android.content.Context;
+import android.text.TextUtils;
+import android.util.Base64;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import com.android.volley.AuthFailureError;
 import com.android.volley.Header;
 import com.android.volley.Request;
 import com.android.volley.RequestTask;
+import com.android.volley.VolleyLog;
 import com.android.volley.toolbox.AsyncHttpStack;
 import com.android.volley.toolbox.ByteArrayPool;
 import com.android.volley.toolbox.HttpResponse;
 import com.android.volley.toolbox.PoolingByteArrayOutputStream;
 import com.android.volley.toolbox.UrlRewriter;
 import java.io.IOException;
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -244,27 +272,23 @@ private void setHttpMethod(Request<?> request, UrlRequest.Builder builder)
...
      * @param additionalHeaders for the UrlRequest
-     * @param builder that we are adding the request headers to
+     * @param requestParameters parameters that we are adding the request headers to
      * @throws AuthFailureError is thrown if Request#getHeaders throws ones
      */
     private void setRequestHeaders(
-            Request<?> request, Map<String, String> additionalHeaders, UrlRequest.Builder builder)
+            Request<?> request,
+            Map<String, String> additionalHeaders,
+            CurlLoggedRequestParameters requestParameters)
```
jbir789, line 281: It's really a nitpick, but since we're adding to the requestParameters, I would suggest moving it as the first method parameter (same for other similar methods)

jbir789 (approved):

src/main/java/com/android/volley/cronet/CronetHttpStack.java
jpd236: Done

PR #371, opened 2020-09-30
Set Content-Type header for requests with bodies.

jpd236:
This is handled by HurlStack#addBody and should be handled in the
Cronet stack as well.

### DIFF
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -28,6 +28,7 @@ import com.android.volley.RequestTask;
 import com.android.volley.VolleyLog;
 import com.android.volley.toolbox.AsyncHttpStack;
 import com.android.volley.toolbox.ByteArrayPool;
+import com.android.volley.toolbox.HttpHeaderParser;
 import com.android.volley.toolbox.HttpResponse;
 import com.android.volley.toolbox.PoolingByteArrayOutputStream;
 import com.android.volley.toolbox.UrlRewriter;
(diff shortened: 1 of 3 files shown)


src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -287,8 +288,10 @@ private void setRequestHeaders(
 
     /** Sets the UploadDataProvider of the UrlRequest.Builder */
     private void addBodyIfExists(
-            CurlLoggedRequestParameters requestParameters, @Nullable byte[] body) {
-        requestParameters.setBody(body);
+            CurlLoggedRequestParameters requestParameters,
+            Request<?> request,
+            @Nullable byte[] body) {
```
jbir789, line 293: What do you think of just passing the body content type, since we don't use the full request for anything else ? (given that the body can differ depending on the http method)

jbir789 (approved):

src/main/java/com/android/volley/cronet/CronetHttpStack.java
jpd236: Done. I was thinking that we should avoid invoking the getBodyContentType method if the body was null since there's no real guarantee that the method works in that case, but I can't imagine a situation where someone would have a request without a body and somehow do something bad in getBodyContentType.

This also helps resolve some unease I had about using getBodyContentType instead of getPostBodyContentType for DEPRECATED_GET_OR_POST requests, which HttpClientStack is doing. HurlStack doesn't, but I think that's probably an oversight.

PR #372, opened 2020-09-30
Make it possible to extend classes created with Builders.

jpd236:
AsyncRequestQueue, BasicAsyncNetwork, and CronetHttpStack all use a new Builder
pattern to simplify construction. However, they are also all possible to
subclass by design, just like their synchronous counterparts. At the moment,
this is impossible, since the constructor is private, and the Builder class
does not know about subclasses in order to build them.

So, switch to a pattern that permits both Builders and inheritance. Builder
classes are now genericized and can be extended. Each instantiable class
has both a public abstract Builder class for extending and an internal
Builder class for actual creation, which can be accessed with a new
static builder() method. The work in build() has been moved to the
constructor itself to ensure it is invoked for subclasses.

This is a significantly more verbose pattern, but the only other options
would be to revert back to regular constructors, or to try and disallow
subclassing and provide other hooks (e.g. callback interfaces) where
customization is desired. This feels like the best compromise to keep
Volley mostly as it was before, just with some Builders to help with
instantiation.

### DIFF
--- a/src/main/java/com/android/volley/AsyncRequestQueue.java
+++ b/src/main/java/com/android/volley/AsyncRequestQueue.java
@@ -41,7 +41,9 @@ import java.util.concurrent.TimeUnit;
  * An asynchronous request dispatch queue.
  *
  * <p>Add requests to the queue with {@link #add(Request)}. Once completed, responses will be
- * delivered on the main thread (unless a custom {@link ResponseDelivery} has been provided)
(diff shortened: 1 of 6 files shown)


jpd236:
Actually... let's hold off on this for now. The fact that of many of Volley's classes can be subclassed has led to pain in the past with hard-to-define behavior guarantees. I suspect we can solve most use cases for subclassing by either providing official fakes for test purposes or providing interfaces to hook into specific parts of the request flow. If not, we can always open this door later.

PR #373, opened 2020-10-01
Add hook to allow customization of UrlRequest.Builder.

jpd236:
Can be used (for example) to enable socket tagging or add a
RequestFinishedInfo.Listener to monitor request results, without
needing to make global changes to the CronetEngine.

### DIFF
--- a/src/main/java/com/android/volley/cronet/CronetHttpStack.java
+++ b/src/main/java/com/android/volley/cronet/CronetHttpStack.java
@@ -41,6 +41,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import org.chromium.net.CronetEngine;
 import org.chromium.net.CronetException;
@@ -58,6 +59,7 @@ public class CronetHttpStack extends AsyncHttpStack {
     private final CronetEngine mCronetEngine;
     private final ByteArrayPool mPool;
     private final UrlRewriter mUrlRewriter;
+    private final RequestListener mRequestListener;
 
     // cURL logging support
     private final boolean mCurlLoggingEnabled;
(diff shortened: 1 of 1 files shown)


src/main/java/com/android/volley/cronet/CronetHttpStack.java
```diff
@@ -504,12 +521,40 @@ public void logCurlCommand(String curlCommand) {
...
+
+        /** @see AsyncHttpStack#getNonBlockingExecutor() */
+        protected Executor getNonBlockingExecutor() {
+            return mStack.getNonBlockingExecutor();
+        }
+
+        /** @see AsyncHttpStack#getBlockingExecutor() */
+        protected Executor getBlockingExecutor() {
+            return mStack.getBlockingExecutor();
+        }
```
jbir789, line 555: Why do we need this? Is it so `#onRequestPrepared` can choose to defer its execution to one of these executor? If so, shouldn't that be up to the application to decide which executor to use?

jbir789 (approved):

src/main/java/com/android/volley/cronet/CronetHttpStack.java
jpd236: I expect one common use case here to be calling ExperimentalUriRequest.Builder.setRequestFinishedListener, and subclassing RequestFinishedInfo.Listener requires providing the executor on which callbacks will be invoked later on.

I did think about keeping the interface simpler and just requiring clients to provide their own executors. While this is probably reasonable for clients that plan to provide their own (likely shared) executors as an input to Volley, keep in mind that Volley does provide default constructors with default thread pools to use. So if you're just using the default config, as I expect many applications will do, you have no way of accessing those executors and would have to create / provide a separate one just for this interface. Thus, we expose them here, just like we provide them in getters in AsyncHttpStack.

jbir789: Sounds good, thanks for the extra context

PR #374, opened 2020-10-02
Fix empty responses when content length is unknown.

jpd236:
For whatever reason, it seems that HttpUrlConnection#getContentLength may
return -1 even if the response includes a Content-Length header. This
still worked fine in Volley, which only used this header to try to
pre-size buffers for reading the response. However, this regressed when
we introduced the change to HttpResponse to support byte[] content in
addition to InputStream content, as we used the content-length to determine
whether content was expected.

Resolve this regression by just checking that content was provided (in
either form) before returning it, or else returning null, without
requiring a valid Content-Length header to be set.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/HttpResponse.java
+++ b/src/main/java/com/android/volley/toolbox/HttpResponse.java
@@ -107,13 +107,12 @@ public final class HttpResponse {
      */
     @Nullable
     public final InputStream getContent() {
-        if (mContentLength == -1) {
-            return null;
-        }
         if (mContent != null) {
             return mContent;
-        } else {
+        } else if (mContentBytes != null) {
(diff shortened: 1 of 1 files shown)


jpd236:
One thing to note here - HurlStack (and possibly HttpClientStack - unclear at a glance) explicitly checks for response codes that indicate lack of a response body and will actually ignore the response body in those cases, even if one is provided. In this case, getContent() will always return null, and BasicNetwork will map it to an empty byte array when returning the response (which must be non-null).

In contrast, CronetHttpStack has no special handling of these response codes at the moment. This seems fine enough to me, though - if the response truly has no content, you'll still get an empty byte array passed to HttpResponse, so there's no difference to the application. If it has content... then it seems fine enough to just return it.

jbir789 (approved):

PR #376, opened 2020-10-13
Actually allow applications to provide custom executors.

jpd236:
When the interface was converted to an abstract class, we forgot to make the methods public.

Also add Javadoc to clarify the purpose of the provided queues. Apps may elect to provide shared executors while not making use of these queues for simplicity; this is fine as long as apps are not strongly dependent on the priority scheme. Note that ideally, most work is done on the non-blocking pool, and the priority is not relevant anyway.


jbir789:
Good catch!

jbir789 (approved):

PR #380, opened 2020-10-29
Clean up cache initialization in AsyncRequestQueue.

### LINKED ISSUE #379: AsyncRequestQueue initializes cache on non-blocking thread
Currently, AsyncRequestQueue#start will perform a blocking initialization of the cache on the non-blocking executor. The intent here was to ensure that the cache was initialized before any network operations could progress. But this is flawed for a couple reasons:

1. There's no guarantee that the non-blocking executor will have a single thread, so even if this operation blocks, it's (theoretically) possible for another thread to progress further and try to access the cache before it is initialized.

2. We're blocking on the non-blocking executor, which violates the parameters in which those threads are meant to work. If those pools are shared outside Volley, this may cause suboptimal behavior in the app. It can also violate reasonable StrictMode checks (e.g. preventing disk I/O on non-blocking threads).

We should find another reasonable way to ensure the cache is initialized prior to performing cache operations.

jpd236:
The previous implementation had two key flaws:

- It performed a blocking task in the non-blocking thread pool
- It failed to prevent requests from attempting cache reads prior to
  initialization completing, assuming the relevant thread pool has
  more than one available thread.

To resolve this, we perform initialization asynchronously. Until
it completes, we hold all queued requests in a separate queue.
In practice, initialization should finish quickly, and it should
be rare for this to have any impact on the request flow.

Chained futures would make for a cleaner implementation - we could
store a future tied to initialization and add listeners upon
future completion to begin the request, avoiding the need to track
these requests ourselves and ensure thread safety - but they are
unfortunately not usable in Volley, which only supports Java 7.

Fixes #379

### DIFF
--- a/src/main/java/com/android/volley/AsyncRequestQueue.java
+++ b/src/main/java/com/android/volley/AsyncRequestQueue.java
@@ -25,9 +25,10 @@ import com.android.volley.AsyncCache.OnGetCompleteCallback;
 import com.android.volley.AsyncNetwork.OnRequestComplete;
 import com.android.volley.Cache.Entry;
 import java.net.HttpURLConnection;
+import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.List;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.PriorityBlockingQueue;
@@ -83,6 +84,17 @@ public class AsyncRequestQueue extends RequestQueue {
     private final WaitingRequestManager mWaitingRequestManager = new WaitingRequestManager(this);
 
     /**
+     * Requests which have been queued before cache initialization has completed.
+     *
+     * <p>These requests are kicked off once cache initialization finishes. We avoid enqueuing them
+     * sooner as the cache may not yet be ready.
+     */
+    private final List<Request<?>> mRequestsAwaitingCacheInitialization = new ArrayList<>();
+
+    private volatile boolean mIsCacheInitialized = false;
+    private final Object mCacheInitializationLock = new Object[0];
+
+    /**
      * Sets all the variables, but processing does not begin until {@link #start()} is called.
      *
      * @param cache to use for persisting responses to disk. If an AsyncCache was provided, then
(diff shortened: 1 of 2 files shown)


src/main/java/com/android/volley/AsyncRequestQueue.java
```diff
@@ -119,34 +131,37 @@ public void start() {
...
-        mNonBlockingExecutor.execute(
-                new Runnable() {
-                    @Override
-                    public void run() {
-                        // This is intentionally blocking, because we don't want to process any
-                        // requests until the cache is initialized.
-                        if (mAsyncCache != null) {
-                            final CountDownLatch latch = new CountDownLatch(1);
+        // Kick off cache initialization, which must complete before any requests can be processed.
+        if (mAsyncCache != null) {
```
jbir789, line 135: It's out of scope here, but it would be a bit easier to follow if the notion of initializing cache was abstracted, so `start()` doesn't have to deal directly with checking which cache is in use and just try to initialize "a" cache.

Since it requires a refactor of the class, I don't think we should do it now. But something to keep in mind as a future clean up, when we do further modifications to the async code (like making it thread safe?)

src/test/java/com/android/volley/AsyncRequestQueueTest.java
```diff
@@ -161,4 +143,57 @@ public void removeRequestEventListener_removesListener() throws Exception {
         verifyNoMoreInteractions(listener);
         queue.stop();
     }
+
+    @Test
+    public void requestsQueuedBeforeCacheInitialization_asyncCache() {
+        AsyncCache mockAsyncCache = mock(AsyncCache.class);
+        queue = createRequestQueue(mockAsyncCache);
```
jbir789, line 150: Could `queue` be a local var instead? Since it's not the one set during setup?

jbir789 (approved):

src/main/java/com/android/volley/AsyncRequestQueue.java
jpd236: It could make sense to fork off some helper classes from AsyncRequestQueue for the cache and network layers, similar to how CacheDispatcher / NetworkDispatcher end up abstracting those details away from RequestQueue. I'm hesitant to push it all the way down to the cache layer unless we can find a way to avoid having to duplicate logic in each implementation. (We're somewhat hampered by Cache being an interface rather than an abstract class, which prevents us from being able to expose equivalent initialization methods in both.)

src/test/java/com/android/volley/AsyncRequestQueueTest.java
jpd236: Sure, done

jbir789 (approved):

PR #381, opened 2020-11-09
Migrate from Travis CI to GitHub Actions.

jpd236:
Travis CI is moving to billed plans per https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing. Volley's requirements are simple and should work fine with GitHub Actions, which would hopefully be supported as long as GitHub itself is.


jpd236:
I *think* this will work, but it doesn't seem like the new action is triggering. Not sure if it's an issue with the config, or if the PR must be merged first before the action will run.

.github/workflows/gradle-build.yaml
```diff
@@ -0,0 +1,27 @@
...
+          java-version: 1.8
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Build with Gradle
+        run: ./gradlew --continue verifyGoogleJavaFormat build connectedCheck
+      - name: Publish snapshot
```
joebowbeer, line 23: This should be conditioned on push?

joebowbeer (changes requested):

joebowbeer:
The pull_request action of the target is executed, so this will run for the first time when it is pushed.

.github/workflows/gradle-build.yaml
jpd236: The script itself checks to ensure that it only runs on pushes, so it would be redundant to include in the GitHub config as well. (And my preference in general is to put as little logic in the CI config as possible, since it ties us less tightly to using that particular CI provider).

joebowbeer, line 23:
```suggestion
      - if: github.event_name == 'push'
        name: Publish snapshot
```

joebowbeer: I think it's clearer to add this to config. That way your `on` and your conditions are all in one place. Separation of concerns..

jpd236: OK - no real harm in including it for some safety in depth. Keeping it in the script as well though.

.github/workflows/gradle-build.yaml
```diff
@@ -0,0 +1,28 @@
...
+          java-version: 1.8
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Build with Gradle
+        run: ./gradlew --continue verifyGoogleJavaFormat build connectedCheck
+      - if: github.event_name == 'push'
```
joebowbeer, line 23: I suspect you'll need to swap name and if lines, preserving the leading `-`

jpd236: If only there was a way to test these before merging :)

I don't think it should matter, but I was thinking about it for readability anyway since it makes sense to lead with the name. Done. I'm expecting I'll likely need to follow up with another tweak or two once this is merged anyway.

joebowbeer (approved):

jbir789 (approved):

jpd236:
Looks like we're good to go. https://oss.jfrog.org/artifactory/webapp/#/builds/volley/1604964521436/1604964521368 was published by the action successfully.

https://github.com/google/volley/actions/runs/354897706 is the action run for a push
https://github.com/google/volley/actions/runs/354903087 is the action run for a PR, which demonstrates that changes to the workflow made in the PR do get picked up, and that no snapshots are pushed for PRs

I've rotated the snapshot deployment key to invalidate the previous one, and turned down Travis CI builds.

PR #386, opened 2020-12-22
fix: fixed hurl volley error

CoderHusain:
Volley sends data twice if the connection is slow. The line of code added in HurlStack solves that problem


jpd236:
I don't know how this setting relates to sending data more than once. I'm not aware of HttpURLConnection doing any sort of internal retries. Perhaps this makes the request more likely to succeed or completely time out, and it's Volley's retry layer that's causing the retry?

But in any event, it's not suitable for general use; as noted in [the Javadoc](https://developer.android.com/reference/java/net/HttpURLConnection#setChunkedStreamingMode(int)), not all HTTP servers support this mode. If you feel this is appropriate to use in your app, you can extend HurlStack in your application and override `createConnection` to set this property yourself as you'd like.

PR #393, opened 2021-02-01
Invoke RetryPolicy#retry in the blocking executor.

jpd236:
When using the synchronous RequestQueue, retry will be invoked in the
NetworkDispatcher thread. Clients may be relying on the ability to
perform blocking operations in retry(); a common example would be
clearing auth tokens in the event of an auth error, which may
require disk I/O.

In order to keep the same API contract when using AsyncRequestQueue,
rework NetworkUtility to permit BasicAsyncNetwork to invoke retry in
the blocking executor.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/NetworkUtility.java
+++ b/src/main/java/com/android/volley/toolbox/NetworkUtility.java
@@ -113,30 +113,45 @@ public final class NetworkUtility {
 
     /**
      * Attempts to prepare the request for a retry. If there are no more attempts remaining in the
-     * request's retry policy, a timeout exception is thrown.
+     * request's retry policy, the provided exception is thrown.
+     *
+     * <p>Must be invoked from a background thread, as client implementations of RetryPolicy#retry
+     * may make blocking calls.
      *
      * @param request The request to use.
      */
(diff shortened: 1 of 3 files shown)


src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -173,24 +186,22 @@ static void handleException(
...
                     if (request.shouldRetryServerErrors()) {
-                        attemptRetryOnException(
-                                "server", request, new ServerError(networkResponse));
-                    } else {
-                        throw new ServerError(networkResponse);
+                        return new RetryInfo("server", new ServerError(networkResponse));
                     }
-                } else {
-                    // 3xx? No reason to retry.
                     throw new ServerError(networkResponse);
```
sophiahaim, line 199: Is this needed? You have it again below

sophiahaim (approved):

src/main/java/com/android/volley/toolbox/NetworkUtility.java
jpd236: Nope, just needed to generalize the comment. Done.

PR #405, opened 2021-03-29
Docs

paulsmithkc:
Initial draft of converting the training pages to markdown and revising the "Send a simple request" lesson.

Feedback more than welcome. Willing to convert the other two pages, just wanted to get some feedback first.


jpd236:
Sorry if I didn't make this clear enough in the bug - I don't think we want to migrate the docs to GitHub at this time; we'd like to keep developer.android.com (DAC) docs as the source of truth for the primary developer-facing documentation given that this has long been the home of these docs, and I don't think it's wise for us to have two differing copies of the same info. We can think about whether DAC is the right home for the long-term, but for the immediate term, I don't think we'll be making changes there.

If it would make reviewing proposed changes easier and you'd prefer to do it in this format - one thing you could maybe do would be to separate the conversion into two commits: one which just does a simple conversion to MarkDown for easier editing, and another which contains the actual changes you'd like to propose. I still wouldn't merge the PR directly, but I could review/comment on the differences by viewing the second commit and work on upstreaming these into the DAC docs.

paulsmithkc:
That's fine and completely understable. Where should I commit the initial markdown conversion?

Issue #395, opened 2021-02-08
Nice lib but poorly documented


fpiraneo:
Volley seems to be a nice library for a very basic usage but it seems missing (or not documented) some critical functionalities, like:

1. How can I set customer headers on request?
2. How can I read headers on response from the server?
3. How can I send a plain text or binary payload when making a POST or PUT request?
4. How can I read http status codes (200 = OK, 404 = not found, ... do you mean?)

All this in Kotlin, of course, the language google is pushing!

jpd236:
It seems to me that https://developer.android.com/training/volley/request-custom covers a lot of this, and includes Kotlin snippets, though perhaps not always explicitly.

1. The example GsonRequest demonstrates passing in a Map<String, String> of request headers to the Request constructor which will be included in the request. The mechanism is by returning the headers from getHeaders().
2. The documentation for mentions that NetworkResponse includes the response headers.
3. This doesn't seem to be mentioned explicitly, but you would override Request#getBody() to send a custom payload in the request. (And Request#getBodyContentType to set the Content-Type header).
4. As with 2, NetworkResponse includes the headers.

Will leave this open to track better documentation regarding 3. If you have other concrete suggestions on how that page could have helped you more, let us know.

fpiraneo:
1. I don't want to be compelled to use gson for deserialization; not always I need deserialization; so again: It's possibile to use volley with custom headers? Please provide practical example;
2. Very confused, sorry, very confused;
3. Complete doc MUST mentions explicitly how things works;
4. Maybe a working example will clarify this point.

Again: a three theoretical example are very poor documentation for a product made by Google! I think you can do better with your 135,301 employees!

paulsmithkc:
There a large multitude of errors in the example Java code, which have been there for years.  As far a I can tell there is no mechanism to recommend changes to Android Developers Training pages. If the documentation was stored in GitHub as markdown files we could create Pull Requests to suggest fixes/corrections.

jpd236:
> There a large multitude of errors in the example Java code, which have been there for years. As far a I can tell there is no mechanism to recommend changes to Android Developers Training pages. If the documentation was stored in GitHub as markdown files we could create Pull Requests to suggest fixes/corrections.

That's a fair point, although there would be drawbacks to migrating the existing doc given how long it has been around. I see this more as a limitation of the developer.android.com page that could be more broadly beneficial (ability to suggest edits to docs there).

[...]

PR #408, opened 2021-04-01
Docs 2

paulsmithkc:
Broke the PR into two commits as requested. First commit is the initial conversion to markdown. Second commit is the revised text of the training pages.

See https://github.com/google/volley/issues/395 and https://github.com/google/volley/pull/405 for background.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -1,0 +1,52 @@
+# Volley Overview
+
+Volley is an HTTP library that makes networking for Android apps easier and most importantly, faster. Volley is available on [GitHub](https://github.com/google/volley).
+
+Volley offers the following benefits:
+- Automatic scheduling of network requests.
+- Multiple concurrent network connections.
+- Transparent disk and memory response caching with standard HTTP [cache coherence](https://en.wikipedia.org/wiki/Cache_coherence).
+- Support for request prioritization.
+- Cancellation request API. You can cancel a single request, or you can set blocks or scopes of requests to cancel.
+- Ease of customization, for example, for retry and backoff.
+- Strong ordering that makes it easy to correctly populate your UI with data fetched asynchronously from the network.
+- Debugging and tracing tools.
+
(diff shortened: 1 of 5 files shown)


jpd236:
Thanks!

This is definitely reviewable, but fairly substantial. It will take some time to go through this all and determine what is/isn't suitable for the general documentation.

jpd236:
Update: I've been in touch with the team maintaining the developer.android.com docs, and we are planning to move the docs over to GitHub (with a redirect at DAC for anyone who still goes there) in the near future. Our internal source is similar to Markdown, so we'll be importing from there for the initial docs - that way we're using the original source and hopefully keeping things higher fidelity. Once that's done, we're also hoping they'll be able to help review these proposed doc changes, but you'll likely need to do some rebasing of your edits on top of the new Markdown files since there'll likely be some differences between your generated conversions and our new source files. Just wanted to give a heads up as you may want to hold off on further changes to avoid churn when the new sources are available.

jpd236:
So sorry this ended up taking so long! But the migrated Markdown versions of the doc are now committed here, so you should be able to rebase your changes on top of them so we can review them properly, if you're still interested in contributing at this point.

From taking a closer skim - it looks like a good portion of the minor tweaks to writing / code snippet are probably improvements, though there are a few I'd disagree with. I don't think the advice on avoiding singletons makes sense here - RequestQueues are expensive to spin up/down, and you do not want multiple instances of them in a running application. I'm on the fence about the WorkManager article - on one hand, it's a fairly standard use of WorkManager, which feels out of scope of Volley's documentation, but I can see it being helpful for someone having trouble stitching them together who might fall into the trap of using RequestFuture or otherwise not understanding the async path.

So it might be good to start with just a PR for the minor improvements which should hopefully be less controversial, and we can discuss the WorkManager article as its own piece?

Juanita932 (approved):

Issue #412, opened 2021-05-27
R8 warning: Missing class: org.chromium.net.UrlRequest$Callback


htteng1976:
Hi, I'm getting this issue when integrating volley 1.2.0 into my project.

Add library dependency:

dependencies {
    ...
    implementation 'com.android.volley:volley:1.2.0'
}

Android Studio 4.2 buid warning:

> Task :library:minifyXXXReleaseWithR8
AGPBI: {"kind":"warning","text":"Missing class: org.chromium.net.UrlRequest$Callback","sources":[{}],"tool":"R8"}

Missing class: org.chromium.net.UrlRequest$Callback

Should I have to add google play service cronet manually?

dependencies {
    ...
    implementation 'com.android.volley:volley:1.2.0'
    implementation 'com.google.android.gms:play-services-cronet:17.0.0'
}

jpd236:
Thanks for the report!

Is this just a build warning? If you're not explicitly opting into using the Cronet stack, you should be able to safely ignore this. The classes that are missing won't be used by other HTTP stacks, so it's fine that they're missing. If you *are* opting to use the Cronet stack, then you'd need to add a Cronet dependency (from Play Services or directly).

If this warning is a major issue, we could consider splitting out a separate :volley-cronet target from Volley itself to contain anything that depends on Cronet. But to me, this feels like a fair bit more build complexity to deal with just for a warning in the optimizer. 1.2.0 has been out for over two months and this is the first report/complaint we've gotten about this.

Will leave this open to get more input, but definitely let us know if this is actually breaking builds or if things aren't working when you omit the cronet dependency. You could also add a `-dontwarn org.chromium.**` to your Proguard config to silence the warning, I think (though you would have to watch out for legitimate issues elsewhere).

htteng1976:
Thanks for the quick reply.

It's just a build warning and didn't cause any build break. I will try your recommendataion to silence this warning. Very appreciated.

jpd236:
Seems that there are other complaints about this from an SDK which depends on Volley, and an SDK can't reasonably disable warnings for such a broad package. So we'll have to look into separating out the -cronet artifact into its own library so apps/libraries depending on Volley alone don't run into issues.

jpd236:
[...]

PR #418, opened 2021-06-30
Refactor Volley into a multi-module project.

jpd236:
This splits the library into three Gradle modules:

- core - volley core + toolbox
- cronet - CronetHttpStack
- testing - common testing code

The primary purpose is to split out CronetHttpStack into its own Maven
artifact, com.android.volley:volley-cronet, to prevent applications
which aren't using it from having to either suppress Proguard/D8
warnings or pull in the unnecessary dependency.

The testing module is needed to share TestRequest between unit tests
in both core and cronet. It is not published to Maven.

This PR also removes support for the build.gradle/rules.gradle split
which allowed Gradle projects to depend on Volley by source and
specify their own SDK settings. This is no longer worth supporting,
and any client that wishes to depend on Volley via Gradle should do
so via published Maven artifacts instead.

See #412


settings.gradle
```diff
@@ -0,0 +1,4 @@
+rootProject.name = 'volley'
+include 'core'
+include 'cronet'
+include 'testing'
```
jbir789, line 4: nit: add an empty line at the end of the file to suppress the Github warning

jbir789 (approved):

settings.gradle
jpd236: done

Issue #419, opened 2021-07-01
Allow sending any JSON with JsonArrayRequest and JsonObjectRequest


jpd236:
Prior to #406, JsonArrayRequest required JSONArray request bodies, and JsonObjectRequest required JSONObject request bodies. Nothing should couple the request type and response type, so it's reasonable to want to send JSONObjects and get back JSONArray, or vice versa. However, the new constructors can end up breaking compilation of existing code which depend on Volley, since the request argument is nullable and thus ambiguous if null is provided. For example, code like:

```java
JsonObjectRequest request = new JsonObjectRequest(Method.POST, url, null, listener, errorListener);
```

compiles fine with the current Volley production release, but suddenly fails to compile here since the compiler doesn't know whether this corresponds to the JSONObject or JSONArray constructor.

I don't see a great, backwards-compatible fix for this. Adding a factory method or builder isn't sufficient because subclassing the request is fairly common too (e.g. to populate custom headers). We could just reorder the arguments in one of them, but that would just make for a more confusing API.

So I think for the short term, we should just revert back to what was in place before. Longer term, if/when we make breaking API changes, we can consider improvements here.

The workaround is to extend JsonRequest directly, e.g.:

```java
public class MyJsonObjectRequest extends JsonRequest<JSONObject> {
    public MyJsonObjectRequest(..., JSONArray request, ...) {
        super(..., request != null ? request.toString() : null, ...);
    }

    @Override
    protected Response<JSONObject> parseNetworkResponse(NetworkResponse response) {
        try {
            String jsonString =
                    new String(
                            response.data,
                            HttpHeaderParser.parseCharset(response.headers, PROTOCOL_CHARSET));
            return Response.success(
                    new JSONObject(jsonString), HttpHeaderParser.parseCacheHeaders(response));
        } catch (UnsupportedEncodingException e) {
            return Response.error(new ParseError(e));
        } catch (JSONException je) {
            return Response.error(new ParseError(je));
        }
    }
}
```

or just avoid using these classes entirely - they're not that complex, and you're probably better off using a strongly-typed JSON object instead.

SushmitSingh:
import androidx.annotation.Nullable;
import com.android.volley.NetworkResponse;
import com.android.volley.ParseError;
import com.android.volley.Response;
import com.android.volley.Response.ErrorListener;
import com.android.volley.Response.Listener;
import com.android.volley.toolbox.HttpHeaderParser;
import com.android.volley.toolbox.JsonRequest;

[...]

PR #406, opened 2021-03-31
Allow sending any JSON with JsonArrayRequest & JsonObjectRequest

paulsmithkc:
Adding support for any JSON object/array to be sent as the body of request.

Also adding missing javadoc to the constructors.

PR #420, opened 2021-07-01
Remove new constructors from JsonRequests which are breaking builds.

jpd236:
These conflict with pre-existing constructors - since the request argument is nullable, existing code which passes in null requests is no longer invoking an unambiguous constructor.

See #419 for more details and a workaround.

These constructors were added in #406 and have not yet been included in a public release, so they are safe to remove.

### DIFF
--- a/src/main/java/com/android/volley/toolbox/JsonArrayRequest.java
+++ b/src/main/java/com/android/volley/toolbox/JsonArrayRequest.java
@@ -25,11 +25,10 @@ import com.android.volley.Response.Listener;
 import java.io.UnsupportedEncodingException;
 import org.json.JSONArray;
 import org.json.JSONException;
-import org.json.JSONObject;
 
 /**
  * A request for retrieving a {@link JSONArray} response body at a given URL, allowing for an
- * optional {@link JSONObject} to be passed in as part of the request body.
+ * optional {@link JSONArray} to be passed in as part of the request body.
  */
 public class JsonArrayRequest extends JsonRequest<JSONArray> {
 
@@ -43,30 +42,6 @@ public class JsonArrayRequest extends JsonRequest<JSONArray> {
     public JsonArrayRequest(
             String url, Listener<JSONArray> listener, @Nullable ErrorListener errorListener) {
         super(Method.GET, url, null, listener, errorListener);
-    }
-
-    /**
-     * Creates a new request.
-     *
-     * @param method the HTTP method to use
-     * @param url URL to fetch the JSON from
-     * @param jsonRequest A {@link JSONObject} to post with the request. Null indicates no
-     *     parameters will be posted along with request.
-     * @param listener Listener to receive the JSON response
-     * @param errorListener Error listener, or null to ignore errors.
-     */
-    public JsonArrayRequest(
-            int method,
-            String url,
-            @Nullable JSONObject jsonRequest,
-            Listener<JSONArray> listener,
-            @Nullable ErrorListener errorListener) {
-        super(
-                method,
-                url,
-                jsonRequest != null ? jsonRequest.toString() : null,
-                listener,
-                errorListener);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jpd236:
@paulsmithkc FYI - sorry to have to revert this part of #406; this issue came up when testing the latest version of Volley against Google's internal applications, which means it likely also impacts other applications in the wild. We'll keep the request open for the longer term if/when we decide to make API breaking changes.

paulsmithkc:
In that case the constructors could easily be combined. They would just need to accept a parent class of both `JSONObject` and `JSONArray`.  That would prevent the constructors from being ambiguous.

jpd236:
Unfortunately, both of those classes extend directly from `Object`. So while it wouldn't be ambiguous, it wouldn't offer sufficient type safety to merit the API change.

paulsmithkc:
The problem is that the existing system expects that the return type and the input type of the service will always be the same.

So at present you cannot:
* Send an object and get back an array
* Send an array and get back an object

paulsmithkc:
If you look into the base request class (JsonRequest) you'll see that it accepts the body of the request as a string. And doesn't have any expectations about the content of that string. Which means that in JsonObjectRequest and JsonArrayRequest the only expectation is that body parameter has the toString() method.

Thankfully toString() is defined by Object, so this can be accepted without any issues.

Thus overall, allowing any object to be passed as the body actually doesn't create any type safety issues.

jpd236:
I don't think we want to open the door here for arbitrary objects in the request. It is compile-time safe and will not crash at runtime, but can result in invalid requests because the Content-Type specified in JsonRequest is application/json, but the request cannot be guaranteed to actually be that type.

Arguably this is also a flaw in JsonRequest since it takes an arbitrary String. Overall, I would say these APIs are not great! But the classes themselves are very simple and if you really want to use them, you can use the workaround noted in #419. If/when we decide to make breaking changes, we can clean this up properly and make a coherent set of APIs.

Personally, I'd avoid using these request classes anyway, and would implement a custom Request subclass using a strongly-typed, reflection-free deserialization library like Moshi or Kotlin Serialization, rather than reading directly from a raw JSONObject.

paulsmithkc:
The issue for me here is that these are introduced and recommended in the documentation as the way that you should be sending JSON data with Volley. I know the documentation changes are pending, but this would be an important thing to address when the docs are migrated to Github.

Volley is one the technologies that I've been introducing my students to in my Android course. (College level.) Unfortunately this issue is one that they run into immediately, when they are getting started with Volley and reading the official docs.

jpd236:
Agreed that we should clarify recommendations around how best to make/structure JSON requests in the docs. I'm curious about the scenario that leads to this issue, though - if students are solving a canned problem for academic purposes, wouldn't it be simplest to just use JSONObjects on both sides? If you really wanted to just use an array on either side, you could just wrap it in an object.

(BTW, we haven't forgotten about the pending doc update here)

paulsmithkc:
Let's say your building an ecommerce website. And you want to implement a product search. You need to send a few fields to the server:

* q - the search query
* category
* minPrice
* maxPrice
* sortBy
* page

The back-end for this service is happy to accept these parameters as query parameters, form fields, or a json object.

However the results are returned as an array of products. Which is a perfectly reasonable and sensical thing to do here.

JsonObjectRequest cannot handle this because it expects the results to be an object.

JsonArrayRequest is the obvious answer since the result type is an array. So this is what most students try. But they can't send the data the server, as obviously the server cannot accept these parameters as an array.

* If they're smart, they realize they can include them in the URL. But figuring out how to use UriBuilder in Java is a pain. So they resort to using string concatenation and forget to sanitize the inputs.

* They cannot send the data as form fields using the constructor available. Although occasionally they'll find that they can override getParams() on stack overflow.

* They cannot send the data as a json object, as there is no constructor for it. I have started to advise them on using GSON instead and create their own request subclass. And more of my students get this with the rewritten docs I have them this semester. But it still seems excessive to subclass Request (not an easy thing) for such a simple task.

jbir789 (approved):

jpd236:
If it were up to me, I'd advise students in that situation to return the search response as an object including a result array. It's more future-proof in case there's additional metadata beyond just the result list - such as a status code to indicate that the result of the request, or some debugging information like how long the request took. And it happens to work neatly around the limitation here, since your request and response would both be objects. On the other hand, if you build the service around returning a JSONArray directly, it'd be impossible to add additional metadata to the response later without breaking older clients.

PR #421, opened 2021-07-02
Allowing both objects and arrays to be posted

paulsmithkc:
These changes allow either a JSON object or array to be posted the server.

See #406 and #420


jpd236:
See https://github.com/google/volley/pull/420#issuecomment-873230155 for details on why we don't want to make this part of Volley for now, and #419 for the workaround if you need to post arbitrary JSON strings in a JSON request.

PR #423, opened 2021-08-09
Fix listing of dependencies in Volley POM files.

### LINKED ISSUE #402: Investigate declaration of declaration in generated POM file
Volley's POM file doesn't currently declare any dependencies. This is a bit unexpected, since we have an implementation dependency on androidx.annotations.

This dep has been around in some form since 1.1.1 (see https://github.com/google/volley/commit/a8424005b3ba6d17fc9e123a2176da702f3ead10). At the time we added it, it was added as a "provided" dependency explicitly to avoid it showing up as a requirement for apps which depend on Volley. When we updated to a newer Gradle plugin, we changed this to "implementation" which isn't equivalent. "compileOnly" would be the right change. But since the POM file doesn't end up picking up the deps, it's functionally equivalent.

We should:

1) Fix this dependency declaration. It seems fine to revert it back to compileOnly since calling apps do not need to depend on androidx.annotations to use Volley.

2) Figure out why the deps aren't propagating to the POM file when declared as "implementation", to ensure that any future deps don't get quietly dropped.

Note that androidx.annotation isn't available on Maven Central, just Google Maven. This means that if we wanted to declare it, or any other Google Maven projects, as dependencies, we'd be ineligible to publish future releases on Maven Central. That's definitely a concern.

jpd236:
We were depending directly on the .aar, which loses the dependency information from the build configuration itself.

This caused android.annotations to (correctly) show up as a dependency. Since it is only needed for annotations (like `@Nullable`) that don't need to be retained at runtime, we can mark this as a compileOnly dependency, matching its former "provided" state, to prevent Volley users from needing it. However, this introduces a compiler warning in the tests, since `@RestrictTo` has RetentionPolicy.CLASS and is not present at compile time; since we apply -Werror to all compile tasks, this becomes a compiler error. For now, to keep the build as close as possible to the prior state, add the annotations as an implementation dependency to the tests; a proper fix is tracked in #424.

The only resulting change to the POM files is that volley-cronet declares a dependency on volley.

Note that volley-cronet doesn't declare a dependency on Cronet because there are two plausible choices for how to use Cronet - the version in GmsCore, or the standalone version. There's no great way to codify this dependency choice, so leave it unstated.

Fixes #402

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -5,24 +5,17 @@ android {
 }
 
 dependencies {
-    implementation "androidx.annotation:annotation:1.0.1"
+    compileOnly "androidx.annotation:annotation:1.0.1"
 
     testImplementation project(":testing")
     testImplementation "junit:junit:4.12"
     testImplementation "org.hamcrest:hamcrest-library:1.3"
     testImplementation "org.mockito:mockito-core:2.19.0"
     testImplementation "org.robolectric:robolectric:3.4.2"
+    // TODO(#424): Fix this dependency at the library level.
+    testImplementation "androidx.annotation:annotation:1.0.1"
 }
 
-publishing {
-    publications {
-        library(MavenPublication) {
-            artifactId 'volley'
-            pom {
-                name = 'Volley'
-                description = 'An HTTP library that makes networking for Android apps easier and, most importantly, faster.'
-            }
-            artifact "$buildDir/outputs/aar/core-release.aar"
-        }
-    }
-}
+project.ext.artifactId = 'volley'
+project.ext.pomName = 'Volley'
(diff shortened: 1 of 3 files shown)


jbir789 (approved):

jpd236:
FYI - I had to make a small change to get this to build; by changing the annotation dependency from implementation to compileOnly, the tests broke because of a compiler warning due to a missing annotation referenced in the library class files, and since `-Werror` is on.

For the library itself, this behavior seems to match that of version 1.2.0, where the POM files didn't declare the dependency either, so rather than try to fix it now, I've filed #424 to invest a longer-term solution and worked around it by declaring the annotations as a test-time dependency as well.

PR #453, opened 2023-06-14
Only log debug messages when DEBUG is True

### LINKED ISSUE #452: Potential debug logging even when DEBUG is False
core/src/main/java/com/android/volley/toolbox/NetworkUtility.java logs the message request when the network response is "slow" even when DEBUG is turned off since the conditional statement uses || rather than && when checking whether the message should be logged.

vorlon77:
According to the internal comments and other usage, this message should only be logged when DEBUG is True. The log entry can potentially contain sensitive data in the case where authentication tokens are part of a GET request and so it should be possible for the user of Volley to disable logging by setting DEBUG to False even in the case where the network request is "slow."

Fixes #452

### DIFF
--- a/core/src/main/java/com/android/volley/toolbox/NetworkUtility.java
+++ b/core/src/main/java/com/android/volley/toolbox/NetworkUtility.java
@@ -50,7 +50,7 @@ final class NetworkUtility {
     /** Logs requests that took over SLOW_REQUEST_THRESHOLD_MS to complete. */
     static void logSlowRequests(
             long requestLifetime, Request<?> request, byte[] responseContents, int statusCode) {
-        if (VolleyLog.DEBUG || requestLifetime > SLOW_REQUEST_THRESHOLD_MS) {
+        if (VolleyLog.DEBUG && requestLifetime > SLOW_REQUEST_THRESHOLD_MS) {
             VolleyLog.d(
                     "HTTP response for request=<%s> [lifetime=%d], [size=%s], "
                             + "[rc=%d], [retryCount=%s]",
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -50,7 +50,7 @@ private NetworkUtility() {}
     /** Logs requests that took over SLOW_REQUEST_THRESHOLD_MS to complete. */
     static void logSlowRequests(
```
jpd236, line 51: Let's rename this to "logRequestSummary" and update the Javadoc to reflect, e.g.:

"Logs a summary about the request when debug logging is enabled."

core/src/main/java/com/android/volley/toolbox/NetworkUtility.java
```diff
@@ -50,7 +50,7 @@ private NetworkUtility() {}
...
             long requestLifetime, Request<?> request, byte[] responseContents, int statusCode) {
-        if (VolleyLog.DEBUG || requestLifetime > SLOW_REQUEST_THRESHOLD_MS) {
+        if (VolleyLog.DEBUG && requestLifetime > SLOW_REQUEST_THRESHOLD_MS) {
```
jpd236, line 53: Let's remove the "requestLifetime > SLOW_REQUEST_THRESHOLD_MS" check altogether and just log this when VolleyLog.DEBUG is true.

I think we should retain the behavior that when DEBUG is true, we log all requests, not just slow ones. So we can remove all of the logic around slow requests, including the SLOW_REQUEST_THRESHOLD_MS constant.

jpd236 (changes requested):
Suggesting a few small changes per discussion in #452.

Let me know if you'd rather we take over from here and we can adopt this PR, since the scope has changed a bit from your proposal. Either way is fine with me.

jpd236 (approved):

PR #456, opened 2023-08-03
Set network tag in CronetHttpStack

lucaseckels:
This requires updating the cronet version


jbiral (approved):

jbiral (changes requested):
It looks like the tests are failing, which could be related to the change in Cronet version. Let's address this issue, but otherwise it looks good to me!

jbiral (approved):

Issue #431, opened 2021-11-03
The JsonObjectRequest and JsonArrayRequest don't support 204: No Content


rhbvkleef:
If the server sends a 204 No Content, Volley still attempts to read the message body, and decode it. It shouldn't, or it should permit an empty body somehow.

jpd236:
Since these classes have to return a valid JSONObject/JSONArray object per the API contract, I suppose the question here is whether an empty body should be treated as equivalent to a body containing "{}" (for objects) or "[]" for arrays. The semantics here inherently depend on the application. I could easily see a behavior change here making things worse - if a server starts sending 204s unexpectedly, would it be better to have a client crash due to unexpected response, where the issue might be caught in earlier testing, or silently treat the response as empty, where it could result in data loss (if the app now thinks there is no server-side data available)?

Given how long this class has been around with its existing behavior, I'm hesitant to change it, though I could be convinced. To be honest, though, I'd generally recommend against using these classes either way; JSONObject/JSONArray are low-level, untyped primitives, and you'd likely be better off using an efficient strongly-typed JSON parser like Moshi in a real application. They're also fairly small and easy enough to adapt to suit your needs (e.g. you can always override it and change parseNetworkResponse to check for an empty response before delegating to the regular implementation, if you want to handle it differently.

Will keep this open for now in case there is other feedback.

Zablas:
Ran into this issue as well. It seems to assume that 204 is an error and you can't even check for the status code in the error listener because networkResponse is null. It would be nice to have it handled in the regular listener at least.

jpd236:
I don't think it's assuming that 204 is an error; I believe what's happening is that the parsing logic (e.g. at https://github.com/google/volley/blob/master/core/src/main/java/com/android/volley/toolbox/JsonObjectRequest.java#L103) is trying to parse the empty string as Json and failing to do so, which results in the error callback being invoked with a ParseError.

The workaround is to extend JsonObjectRequest/JsonArrayRequest and override parseNetworkResponse to catch this case however you see fit. For example, something like:

```java
@Override
protected Response<JSONObject> parseNetworkResponse(NetworkResponse response) {
    if (response.statusCode == 204) {
        return Response.success(new JSONObject(), HttpHeaderParser.parseCacheHeaders(response));
    }
    return super.parseNetworkResponse(response);
}

rhbvkleef:
[...]

PR #467, opened 2025-10-22
Fix: Solve 204 No Content Error

jdalvar3z:
### Solve 204 No Content Error

_See issue_ #431 _for Context_

- Refactored parseNetworkResponse to handle 204 No Content responses gracefully. 
- Added check for null or empty response.data to return Response.success with an empty JSONObject instead of triggering a Response.error. 
- Optimized code by extracting HttpHeaderParser.parseCacheHeaders to a local variable for reuse, improving readability and performance. 
- Maintained existing functionality for non-empty responses while enhancing clarity and maintainability.

### DIFF
--- a/core/src/main/java/com/android/volley/toolbox/JsonObjectRequest.java
+++ b/core/src/main/java/com/android/volley/toolbox/JsonObjectRequest.java
@@ -93,7 +93,6 @@ public class JsonObjectRequest extends JsonRequest<JSONObject> {
     protected Response<JSONObject> parseNetworkResponse(NetworkResponse response) {
         Cache.Entry cacheHeaders = HttpHeaderParser.parseCacheHeaders(response);
         try {
-            //Check Empty Response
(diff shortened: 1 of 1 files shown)


jpd236:
My stance here is largely unchanged from what I said here: https://github.com/google/volley/issues/431#issuecomment-959709999

In particular, we're making an assumption that an empty response body should be treated as equivalent to "{}", and while that seems logical, I can easily imagine existing applications that would not expect that change and could fail in significant ways. While I agree this would probably have made sense if this were a brand new class, I'm very hesitant to change the existing behavior of a long-standing class in a semantically meaningful way.

Our priority at this point is maintaining Volley's functionality for existing users. New users shouldn't be using these classes as noted there - or, for that matter, Volley in general. There are more modern recommendations at https://developer.android.com/develop/connectivity/network-ops/connecting#choose-http-client. So while it may seem counterintuitive, I think it's best to maintain existing behavior in case of any doubt.

PR #474, opened 2026-08-27
Make charset parameter name comparison case-insensitive

rootkiller6788:
## Problem

`HttpHeaderParser.parseCharset()` only matches the exact lowercase `charset` parameter name when parsing a `Content-Type` header:

```java
if (pair[0].equals("charset")) {
    return pair[1];
}
```

HTTP header parameter names are case-insensitive (RFC 7230 section 3.2.6). A server that emits e.g. `Content-Type: text/plain; Charset=UTF-8` or `CHARSET=UTF-8` therefore has its charset silently ignored, and the response body falls back to the default encoding. For `StringRequest` the default is `ISO-8859-1`, so any non-ASCII text (e.g. UTF-8 JSON or localized content) is decoded as mojibake.

## Fix

Compare the parameter name with `equalsIgnoreCase` so `charset`, `Charset`, and `CHARSET` are all recognized, matching the case-insensitivity of the rest of header handling in Volley (header names are already looked up case-insensitively via a `TreeMap` with `String.CASE_INSENSITIVE_ORDER`).

## Verification

- Reproduced the bug against the unmodified source with a standalone harness: `Content-Type: text/plain; Charset=UTF-8` returned `ISO-8859-1`; after the fix it returns `UTF-8`. `CHARSET=Shift_JIS` is also recognized.
- Added test coverage to `HttpHeaderParserTest.parseCharset()` for mixed- and upper-case `charset` parameter names.


jpd236:
RFC 7230 section 3.2.6 doesn't say anything about the case-sensitivity of "charset", as far as I can tell. But https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1-5 makes it fairly clear that we should compare ignoring case for this, and this feels reasonably low risk.

